PR #3, opened 2024-02-21
Add inference utils

yeandy:
Adding inference utils. PTAL @mailvijayasingh @entrpn

### DIFF
--- a/src/maxdiffusion/utils/inference_utils.py
+++ b/src/maxdiffusion/utils/inference_utils.py
@@ -1,0 +1,382 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Utilities for JAX inference."""
+
+from typing import Any, Callable
+import functools
+
+from flax import core
+from flax import linen as nn
+from flax import struct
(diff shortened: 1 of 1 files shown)


src/maxdiffusion/utils/inference_utils.py
```diff
@@ -0,0 +1,382 @@
...
+import jax.numpy as jnp
+import jax
+import numpy as np
+
+from maxdiffusion import max_logging
+
+FrozenDict = core.frozen_dict.FrozenDict
+
+
+def loop_body(
```
entrpn, line 38: a lot of these functions seem to be copied over from generate_sdxl.py. Can we refactor generate_sdxl so we don't have duplicate code in two files?

src/maxdiffusion/utils/inference_utils.py
```diff
@@ -0,0 +1,382 @@
...
+        return_tensors="np",
+    )
+    inputs.append(text_inputs.input_ids)
+  inputs = jnp.stack(inputs, axis=1)
+  return inputs
+
+
+def tokenize_one(prompt, tokenizer):
+  inputs = []
+  # for _tokenizer in [pipeline.tokenizer, pipeline.tokenizer_2]:
```
entrpn, line 118: remove commented code.

src/maxdiffusion/utils/inference_utils.py
```diff
@@ -0,0 +1,382 @@
...
+def vae_decode(latents, state, pipeline):
+  latents = 1 / pipeline.vae.config.scaling_factor * latents
+  image = pipeline.vae.apply(
+      {"params": state.params}, latents, method=pipeline.vae.decode
+  ).sample
+  image = (image / 2 + 0.5).clip(0, 1).transpose(0, 2, 3, 1)
+  return image
+
+
+def create_device_mesh(config, devices=None, logging=True):
```
entrpn, line 138: same as above, this is duplicated code from max_utils. Can we max_utils functions as imports.

yeandy:
Thanks @entrpn. Synced with Vijaya, and I think we actually don't need a merge right now; just this `inference_utils` branch to use for reference.

PR #4, opened 2024-02-21
cleanup after tests to prevent oom

entrpn:
Some tests download datasets and parses through them which can fill the device memory. This PR deletes dataset files after running those tests.

### DIFF
--- a/src/maxdiffusion/tests/input_pipeline_interface_test.py
+++ b/src/maxdiffusion/tests/input_pipeline_interface_test.py
@@ -14,6 +14,8 @@
  limitations under the License.
 """
 import os
+import pathlib 
+import shutil
 import unittest
 from absl.testing import absltest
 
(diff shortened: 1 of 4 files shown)


src/maxdiffusion/tests/input_pipeline_interface_test.py
```diff
@@ -25,7 +27,12 @@
 from maxdiffusion.input_pipeline.input_pipeline_interface import make_pokemon_train_iterator
 from maxdiffusion import FlaxStableDiffusionPipeline
 
+HOME_DIR = pathlib.Path.home()
 THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+DATASET_DIR = str(HOME_DIR / ".cache" / "huggingface" / "datasets")
+
+def cleanup(output_dir):
+  shutil.rmtree(output_dir)
```
sshahrokhi, line 35: Just a quick check, should we first make sure the output_dir exists and then remove it? I don't think we want the program to fail if the path does not exist, or do we?

PR #8, opened 2024-03-14
Sdxl lightning

entrpn:
Implementation of sdxl lightning inference. Do a better job at clearing TPU memory when converting weights to bf16.


mkmomar72 (approved):
LGTM

PR #16, opened 2024-04-17
Add gradient norm to TensorBoard telemetry

gobbleturk:


entrpn:
This was already added as part of another commit.

PR #19, opened 2024-04-18
Flax/HuggingFace-Based CLIP Score Calculation

anfals:


sshahrokhi (approved):
Thanks Anfal for the help! LGTM! :)

PR #24, opened 2024-04-19
All fid

sshahrokhi:
making the fid scores closer to pytorch fid scores


suexu1025 (approved):
Thanks for including pytorch implementation and matching FID with pytorch's version.

PR #25, opened 2024-04-22
Clip Score w/Batching

anfals:
This is slower in my testing than the sequential clip score in Jax, but faster than pytorch. Will investigate further.

### DIFF
--- a/src/maxdiffusion/metrics/clip/clip_encoder.py
+++ b/src/maxdiffusion/metrics/clip/clip_encoder.py
@@ -80,7 +80,13 @@ class CLIPEncoderFlax:
         image_embeddings /= jnp.linalg.norm(image_embeddings, axis=-1, keepdims=True)
         text_embeddings /= jnp.linalg.norm(text_embeddings, axis=-1, keepdims=True)
 
-        return jnp.tensordot(image_embeddings, text_embeddings)
+        image_embeddings = jnp.expand_dims(image_embeddings, 1)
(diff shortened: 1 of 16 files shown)


src/maxdiffusion/metrics/clip/clip_encoder.py
```diff
@@ -52,12 +56,43 @@ class CLIPEncoderFlax:
...
+        return CLIPEncoderFlax._calculate_clip_score(text_embeddings, image_embeddings, self.logit_scale)
 
-        return outputs.logits_per_image / 100
+    @staticmethod
+    @jax.jit
+    def _calculate_clip_score(text_embeddings, image_embeddings, logit_scale):
+        image_embeddings /= jnp.linalg.norm(image_embeddings, axis=-1, keepdims=True)
+        text_embeddings /= jnp.linalg.norm(text_embeddings, axis=-1, keepdims=True)
+
+        image_embeddings = jnp.expand_dims(image_embeddings, 1)
```
anfals, line 81: Frustratingly, there seems to be a precision issue between Jax matmul vs jax sum and element-wise multiplication. If I went with the latter, with simpler code not requiring this batch_matmul, the difference in scores between PyTorch implementation above and Jax sometimes exceeded 1e-3. Assuming we need to keep within that margin (and that the pytorch implementation we had was the ground truth), I switched to this.

sshahrokhi (approved):
Thanks, LGTM!

PR #27, opened 2024-04-23
Add mllog for inference, eval, and report_end script

ZhiyuLi-goog:
* add mllog for inferencing and evaluation
* a `report_end.py` script to find out the earliest convergence step and print it to mllog
*  an e2e `xpk/v5p-128_e2e_mlperf.sh` to train, eval for all checkpoints, process and generate finall mllog
* tf.io is compatible to both local and gcs path, so we don't the additional step to upload/download npz and tsv

### DIFF
--- a/src/maxdiffusion/models/train.py
+++ b/src/maxdiffusion/models/train.py
@@ -435,7 +435,7 @@ def train(config):
     mllog_utils.train_init_print(config)
     mllog_utils.train_init_stop()
     mllog_utils.train_run_start()
-    mllog_utils.train_step_start(start_step)
+    mllog_utils.train_step_start(start_step, samples_count=0)
     # for checkpointing
     for step in np.arange(start_step, config.max_train_steps):
         example_batch = load_next_batch(data_iterator, example_batch, config)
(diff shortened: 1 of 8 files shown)


entrpn (approved):

src/maxdiffusion/models/train.py
```diff
@@ -448,15 +448,18 @@ def compute_loss(unet_params):
         record_scalar_metrics(train_metric, new_time - last_step_completion, per_device_tflops, learning_rate_scheduler(step))
         write_metrics(writer, local_metrics_file, running_gcs_metrics, train_metric, step, config)
         last_step_completion = new_time
-        if step != 0 and (total_train_batch_size * step) % config.checkpoint_every == 0:
```
ZhiyuLi-goog, line 451: @entrpn Hi Juan, could you check this part?
I think the `step` refers to the index starting from 0.
And we should add one to get `num_step` to calculate `consumed_samples`

entrpn: That makes sense!

PR #29, opened 2024-04-23
Revert "Clip Score w/Batching"

anfals:
I can't get the batched implementation to be faster than just doing sequential and it may introduce some regressions. Let's keep sequential.

### DIFF
--- a/src/maxdiffusion/eval.py
+++ b/src/maxdiffusion/eval.py
@@ -49,6 +49,19 @@ def load_stats(file_path):
     mu = images_data['mu']
     return sigma, mu
 
+
+def calculate_clip(images, prompts):
+    clip_encoder = CLIPEncoderFlax()
+    
+    clip_scores = []
+    for i in tqdm(range(0, len(images))):
+        score = clip_encoder.get_clip_score(prompts[i], images[i])
+        clip_scores.append(score)
+        
+
+    overall_clip_score = jnp.mean(jnp.stack(clip_scores))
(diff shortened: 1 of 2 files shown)


src/maxdiffusion/eval.py
```diff
@@ -74,8 +87,7 @@ def eval_scores(config, images_directory=None):
     captions_df = load_captions(config.caption_coco_file)
```
anfals, line 87: @jfacevedo-google This is the usage of CLIP that's been here for a while. Not sure if the usage of it is wrong or whether some debug code snuck in instead

entrpn: It might be more prudent to parse the caption_coco_file for the ids and match the images and captions with that id, instead of looping over both lists and hoping they are both in order.

anfals: I'll merge this revert in and see if I can quickly play around with this.

anfals: @entrpn Ah see https://github.com/google/maxdiffusion/blob/a8b7bf018a4085aae84ac452dd5105d57f62bf73/src/maxdiffusion/eval.py#L71. That's already what's happening - `load_images` loads the images and then queries the captions_df for the corresponding prompt based on the prompt id. The returned lists are in order already. 

Let me know if you encounter the same issues - i can find some time to debug this further.

entrpn: Sounds good, thank you!

PR #30, opened 2024-04-23
Mlperf jfacevedo training tricks

entrpn:
- Implement v-parameterization
- Implement noise offset
- Implement noise peturbation
- Log FID/CLIP scores to tensorboard
- Make v-parameterization, noise offset, noise peturbation, training/inference schedulers configurable through config.

### DIFF
--- a/src/maxdiffusion/configs/base_2_base.yml
+++ b/src/maxdiffusion/configs/base_2_base.yml
@@ -165,8 +165,10 @@ negative_prompt: ""
 # https://github.com/mlcommons/training/blob/master/stable_diffusion/configs/train_01x08x08.yaml#L26
 guidance_scale: 8
 num_inference_steps: 50
+rescale_zero_terminal_snr: False
 # For preprocessing data to tfrecords
 data_files_pattern: ""
 extracted_files_dir: ""
 tfrecords_dir: ""
 no_records_per_shard: 1000
+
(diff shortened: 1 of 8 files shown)


ZhiyuLi-goog:
Thank you @entrpn. It looks good to me thank you for addition!

I am confident noise augmentation (noise offset, noise peturbation) and schedule would be helpful for convergence. Just curious if these training tricks are expected to use in mlperf submission.  

I think the LR schedule is limited to [warm_up_to_constant](https://github.com/mlcommons/training_results_v3.1/blob/main/NVIDIA/benchmarks/stable_diffusion/implementations/pytorch/conf/sd2_mlperf_train_moments.yaml#L193C13-L193C29) in mlperf [according to the rule](https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc) as you can find `opt_learning_rate_warmup_steps` in Hyperparameters of SD. We may need to double check if noise offset and noise peturbation are allowed as well.

entrpn:
> Thank you @entrpn. It looks good to me thank you for addition!
> 
> I am confident noise augmentation (noise offset, noise peturbation) and schedule would be helpful for convergence. Just curious if these training tricks are expected to use in mlperf submission.
> 
> I think the LR schedule is limited to [warm_up_to_constant](https://github.com/mlcommons/training_results_v3.1/blob/main/NVIDIA/benchmarks/stable_diffusion/implementations/pytorch/conf/sd2_mlperf_train_moments.yaml#L193C13-L193C29) in mlperf [according to the rule](https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc) as you can find `opt_learning_rate_warmup_steps` in Hyperparameters of SD. We may need to double check if noise offset and noise peturbation are allowed as well.

Yes I agree, so they are configurable. Right now I'm leaving them off.

ZhiyuLi-goog (approved):

PR #35, opened 2024-04-25
Mlperf jfacevedo training tricks 2

entrpn:


src/maxdiffusion/max_utils.py
```diff
@@ -511,6 +515,9 @@ def save_checkpoint(pipeline, params, unet_state, noise_scheduler, config, outpu
   )
   if jax.process_index() == 0:
     walk_and_upload_blobs(config, local_output_dir)
+  
+  # delete files in output_dir to save space
+  shutil.rmtree(local_output_dir)
```
ZhiyuLi-goog, line 520: Just curious if `save_pretrained` will be performed on all processes and we need to delete it on all processes as well.

entrpn: `save_pretrained` will be saved on all processes, but uploading the weights is only done in the first process. Then we delete the weights from all processes to save memory. This is mainly to ensure we don't run out of disk memory if checkpoint_every is creating a lot of checkpoints.

src/maxdiffusion/models/train.py
```diff
@@ -451,16 +496,14 @@ def compute_loss(unet_params):
...
         samples_count = total_train_batch_size * step_num
         if step != 0 and samples_count % config.checkpoint_every == 0:
-           if config.eval_at_checkpoint:
-              eval_at_checkpoint(config,
-                   f"{str(step * total_train_batch_size)}",
-                   unet_state, unet_state_mesh_shardings,
-                   vae_state,
-                   vae_state_mesh_shardings,
-                   pipeline, params, mesh, rng)
-           checkpoint_name = f"{step_num=}-{samples_count=}"
```
ZhiyuLi-goog, line 461: This change will break the later evaluation 
* extract step_name and samples count info from the name https://github.com/google/maxdiffusion/blob/mlperf_4/src/maxdiffusion/eval.py#L87
Is it possible to keep the `checkpoint_name` format?

ZhiyuLi-goog (commented):
All looks great except some concerns about the `checkpoint_name` format change which will break the following info extraction.

ZhiyuLi-goog (approved):

PR #37, opened 2024-04-30
unify batch sharding across train/eval

suexu1025:
- unify batch sharding across train/eval

### DIFF
--- a/src/maxdiffusion/multihost_dataloading.py
+++ b/src/maxdiffusion/multihost_dataloading.py
@@ -21,7 +21,7 @@ Adapted from Sholto's:
 https://github.com/sholtodouglas/multihost_dataloading
 """
 from functools import partial  # pylint: disable=g-importing-member
-from typing import Callable
+from typing import Callable, Any, Union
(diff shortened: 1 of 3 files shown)


src/maxdiffusion/multihost_dataloading.py
```diff
@@ -104,3 +104,30 @@ def get_next_batch_sharded(
   input_gdas = jtu.tree_map_with_path(partial(_form_global_array, global_mesh = global_mesh), local_data)
 
   return input_gdas
+
+def get_data_sharded(
```
entrpn, line 108: @suexu1025 is it possible to reuse the same code from [multihost_dataloading.py](https://github.com/google/maxdiffusion/blob/mlperf_4/src/maxdiffusion/multihost_dataloading.py)

suexu1025: The existing can only take tf.data.Dataset.  I can just replace it with `jtu.tree_map_with_path(partial(_form_global_array, global_mesh = global_mesh), local_data)`

entrpn (approved):

PR #38, opened 2024-04-30
Mlperf 4 rescale snr param

entrpn:


sshahrokhi (approved):
LGTM! I assume you have checked already that the run works with these changes?

PR #43, opened 2024-05-03
[MLLOG] fix timestamp of checkpoint saving

ZhiyuLi-goog:


suexu1025 (approved):
Thanks Zhiyu, looks good

PR #45, opened 2024-05-05
Mlperf 4 lr init value

entrpn:
- Config setting for initial lr value
- Add consine learning rate scheduler


suexu1025 (approved):
Thanks! LGTM

PR #48, opened 2024-05-07
precompile train step

entrpn:


ZhiyuLi-goog:
Oh is it the same as this change https://github.com/google/maxdiffusion/pull/44/files?

entrpn:
> Oh is it the same as this change https://github.com/google/maxdiffusion/pull/44/files?

Ah it looks like @suexu1025 implemented this also on her branch. We can discuss which one we want to use.

PR #49, opened 2024-05-07
Mlperf train eval

entrpn:


src/maxdiffusion/models/train.py
```diff
@@ -505,12 +505,13 @@ def compute_loss(unet_params):
                     vae_state_mesh_shardings,
                     pipeline, params, train_metric,
                     checkpoint_name)
-            max_utils.save_checkpoint(pipeline,
+            
+            eval_checkpoints.append(max_utils.save_checkpoint(pipeline,
```
suexu1025: can we add a start_step_to_checkpoint?

suexu1025 (commented):
Thanks @jfacevedo-google! left a bit comment

src/maxdiffusion/input_pipeline/input_pipeline_interface.py
```diff
@@ -98,7 +98,6 @@ def create_dict(moments, input_ids):
       .map(prepare_sample, num_parallel_calls=AUTOTUNE)
       .map(lambda x, y: tf.py_function(partial_tokenize, inp=[x, y], Tout=(tf.float32, tf.float32)), num_parallel_calls=AUTOTUNE)
       .map(create_dict, num_parallel_calls=AUTOTUNE)
-      .shuffle(global_batch_size * 10, seed=config.seed)
       .batch(global_batch_size // jax.process_count(), drop_remainder=False)
```
suexu1025: This might still affect convergence, I got worse result. maybe add it at later PR?

src/maxdiffusion/models/train.py
```diff
@@ -522,6 +523,22 @@ def compute_loss(unet_params):
         mllog_utils.maybe_train_step_log(config, start_step, step_num, samples_count, train_metric)
     max_utils.close_summary_writer(writer)
 
+    del pipeline
+    del params
+    del unet_state
+    del vae_state
+
```
ZhiyuLi-goog, line 530: LGTM. And feel free to merge in.

Just curious how garbage/memory collection is performed in tpu side. We can do something like `gc.collect()` in cpu. Not a problem as long as we don't hit the hbm limit.

ZhiyuLi-goog (approved):

src/maxdiffusion/models/train.py
entrpn: I didn't see any issues with memory but if we run into it, there is a function in max_utils, `delete_pytree` that can explicitly delete memory allocations on TPU.

suexu1025 (approved):

PR #50, opened 2024-05-08
Mlperf train eval 2

entrpn:


src/maxdiffusion/models/train.py
```diff
@@ -531,6 +530,23 @@ def compute_loss(unet_params):
...
 
+    del pipeline
+    del params
+    del unet_state
+    del vae_state
+
+    config.train_new_unet = False
+    for checkpoint in eval_checkpoints:
+        config.unet_checkpoint = checkpoint
+        checkpoint_name = checkpoint.split("/")[-1]
```
ZhiyuLi-goog, line 541: will it be an issue if checkpoint has '/' at the end?
can we do something like

```
os.path.basename(checkpoint.rstrip("/"))
```

ZhiyuLi-goog (approved):
LGTM

suexu1025 (approved):

PR #58, opened 2024-05-13
fix unit tests in main

entrpn:
- Update deprecated calls to fix unit tests.


src/maxdiffusion/max_utils.py
```diff
@@ -306,7 +306,7 @@ def setup_initial_state(model, tx, config, mesh, model_params, unboxed_abstract_
                                                   mesh,
                                                   state_mesh_annotations)
 
-    state_mesh_shardings = jax.tree_map(
+    state_mesh_shardings = jax.tree_util.tree_map(
```
mailvijayasingh, line 309: how is tree_util.tree_map different from tree_map?

mailvijayasingh (approved):

src/maxdiffusion/max_utils.py
entrpn: jax.tree_map is deprecated and aliased, see https://jax.readthedocs.io/en/latest/jax.tree.html

PR #59, opened 2024-05-13
add new pokemon dataset for training. make mllogs optional.

entrpn:
* Add new Pokemon dataset for training.
* Make mllog logs optional (off by default).


src/maxdiffusion/configs/base21.yml
```diff
@@ -108,7 +108,6 @@ max_train_samples: -1
 max_train_steps: 800
 seed: 0
 output_dir: 'sd-model-finetuned'
-tensorboard_dir: 'gs://shahrokhi-maxdiffusion-v5'
```
mailvijayasingh, line 111: shall we rename the bucket?

mailvijayasingh (approved):

PR #61, opened 2024-05-16
Sdxl training

entrpn:


src/maxdiffusion/configs/base_xl.yml
```diff
@@ -79,7 +79,7 @@ ici_tensor_parallelism: 1
 
 # Dataset
 # Replace with dataset path or train_data_dir. One has to be set.
-dataset_name: ''
+dataset_name: 'diffusers/pokemon-gpt4-captions'
```
mailvijayasingh, line 82: Do we want to separate generate and train config files?

src/maxdiffusion/generate_sdxl.py
```diff
@@ -132,14 +129,28 @@ def run(config):
     mesh=mesh
   )
 
+  # if this checkpoint was trained with maxdiffusion
+  # the training scheduler was saved with it, switch it
+  # to a Euler scheduler
+  if isinstance(pipeline.scheduler, FlaxDDPMScheduler):
```
mailvijayasingh, line 135: how does this check if the checkpoint is loaded from `maxdiffusion based training`?

src/maxdiffusion/generate_sdxl.py
```diff
@@ -132,14 +129,28 @@ def run(config):
...
+    params["scheduler"] = noise_scheduler_state
+
   if config.lightning_repo:
     pipeline, params = load_sdxllightning_unet(config, pipeline, params)
 
   scheduler_state = params.pop("scheduler")
   old_params = params
   params = jax.tree_util.tree_map(lambda x: x.astype(weight_dtype), old_params)
   params["scheduler"] = scheduler_state
-  delete_pytree(old_params)
```
mailvijayasingh, line 142: why remove `delete_pytree(old_params)`?

src/maxdiffusion/max_utils.py
```diff
@@ -433,3 +433,53 @@ def get_memory_allocations():
     m_stats = device.memory_stats()
     max_logging.log(f'device : {device.process_index},'
                     f'bytes in use: {m_stats["bytes_in_use"] / gb} / {m_stats["bytes_limit"] / gb} GB')
+
+def calculate_training_tflops(pipeline, unet_params, config):
```
mailvijayasingh, line 437: how about having an argument in the function - `training`/`inference`? and also have an argument about factors like 3 and 3.2?

src/maxdiffusion/configs/base_xl.yml
entrpn: Its getting difficult to manage all the different config files so I'm trying to consolidate them all into single files similar to how maxtext is doing it.

src/maxdiffusion/generate_sdxl.py
entrpn: DDPM scheduler is used for training and it should not be used for inference as it will result in noisy images. More broadly this ensures that someone doesn't accidentally use the wrong scheduler.

src/maxdiffusion/generate_sdxl.py
entrpn: delete pytree was originally being used when some weights are loaded from pt, but now this logic is changed to keep pt -> jax weights in cpu before being moved to host and deleting the pytree is not needed anymore.

src/maxdiffusion/max_utils.py
entrpn: will do it in another commit.

mailvijayasingh (dismissed):

src/maxdiffusion/train_sdxl.py
```diff
@@ -0,0 +1,513 @@
...
+      )
+      writer.flush()
+
+def calculate_num_params_from_pytree(params):
+  params_sizes = jax.tree_util.tree_map(jax.numpy.size, params)
+  total_parameters = jax.tree_util.tree_reduce(lambda x, y: x + y, params_sizes)
+  return total_parameters
+
+def get_params_to_save(params):
+    return jax.device_get(jax.tree_util.tree_map(lambda x: x, params))
```
sshahrokhi: Up until here, can be used for both SD 2 training and SDXL training, can we move them to another file/ appropriate existing files and use them instead of copy pasting?

sshahrokhi (approved):
LGTM! Thanks!

PR #67, opened 2024-05-29
fix  tflop calculations.

### LINKED ISSUE #65: latents concatenation in training flops calculation
https://github.com/google/maxdiffusion/blob/main/src/maxdiffusion/max_utils.py#L452
While training, is it needed to concat latents? 

Should it be concatenated only during inference and not during training?

entrpn:
Fixes #65


mailvijayasingh (approved):
Thanks for fixing this, Juan!

PR #70, opened 2024-06-04
Training optimizations

entrpn:
* Add snr gamma
* Add AOT
* Add timestep bias


src/maxdiffusion/models/train.py
```diff
@@ -381,9 +419,10 @@ def compute_loss(unet_params):
 
 def main(argv: Sequence[str]) -> None:
     max_logging.log(f"Found {jax.device_count()} devices.")
-    cc.set_cache_dir(os.path.expanduser("~/jax_cache"))
     pyconfig.initialize(argv)
     config = pyconfig.config
+    if len(config.cache_dir) > 0:
```
suexu1025, line 424: Is this check safe? Would cache_dir == NONE?

src/maxdiffusion/models/train.py
```diff
@@ -276,21 +305,30 @@ def compute_loss(unet_params):
...
+    with mesh, nn_partitioning.axis_rules(config.logical_axis_rules):
         p_train_step = jax.jit(
             partial(train_step, cache_latents_text_encoder_outputs=config.cache_latents_text_encoder_outputs),
             in_shardings=(unet_state_mesh_shardings, vae_state_mesh_shardings, my_data_sharding, None),
             out_shardings=(unet_state_mesh_shardings, None, None),
             donate_argnums=(0,)
         )
+        max_logging.log("Precompiling...")
+        s = time.time()
+        dummy_batch = get_shaped_batch(config, pipeline)
```
suexu1025, line 318: Do we need flag to turn off pre-compilcation in case the training  data format is different from what we defined?

suexu1025 (approved):
Thanks, looks great, few comments

src/maxdiffusion/models/train.py
entrpn: The way I understand pyconfig works, it still requires the variable name, but with an empty value. if the variable name is not in the config, then an error will be thrown instead, thus using len > 0 instead of None.

src/maxdiffusion/models/train.py
entrpn: I gave it some thought and found it that we should enforce best practices instead of providing users too many options. At the end, the training and inference scripts are reference implementations of a task and it should be expected for a user to modify or create new scripts for their corner cases.

PR #71, opened 2024-06-04
Qinwen/mlperf 4 data

suexu1025:
* step time reduced from 20% for using preprocessed text-embeddeding.


src/maxdiffusion/input_pipeline/input_pipeline_interface.py
```diff
@@ -233,8 +232,8 @@ def get_shaped_batch(config, pipeline):
         config.resolution // vae_scale_factor,
         config.resolution // vae_scale_factor, 8)
   #bs, encoder_input, seq_length
-  batch_ids_shape = (total_train_batch_size, pipeline.text_encoder.config.max_position_embeddings)
+  batch_ids_shape = (total_train_batch_size, pipeline.text_encoder.config.max_position_embeddings, 1024)
```
entrpn: I believe 1024 is the hidden size. See [here](https://huggingface.co/stabilityai/stable-diffusion-2-base/blob/main/text_encoder/config.json#L11)

entrpn (approved):

src/maxdiffusion/input_pipeline/input_pipeline_interface.py
suexu1025: Thanks, updated

PR #81, opened 2024-06-17
Improve logging to remove data input overhead

aireenmei:
* Switch to write_metrics (logging with buffer)
* add log_period. Usage suggestiong: logging will always trigger sync in jax, and break the overlapping of data input and computation. We can set log_period to a small value for debugging, but for submission, we should set to a big value, or even turn it off, only outputs at the end.
* cloud log for training 5000 steps with log_period 100: https://cloudlogging.app.goo.gl/i3a8d4LfTvzrsWqh7
* xprof: https://xprof.corp.google.com/trace_viewer/aireenmei-15432370690143073139?hosts=gke-tpu-fc96cd7e-vn72&host_index=0

### DIFF
--- a/src/maxdiffusion/models/train.py
+++ b/src/maxdiffusion/models/train.py
@@ -149,10 +149,16 @@ def validate_train_config(config):
 
   assert config.max_train_steps > 0 or config.num_train_epochs > 0, "You must set steps or learning_rate_schedule_steps to a positive interger."
 
-def record_scalar_metrics(metrics, step_time_delta, per_device_tflops, lr):
+def record_scalar_metrics(metrics, step_time_delta, num_step, per_device_tflops, lr):
   """Records scalar metrics to be written to tensorboard"""
   metrics['scalar'].update({
-      'perf/step_time_seconds': step_time_delta.total_seconds()
+      'perf/period_time_seconds': step_time_delta.total_seconds()
+  })
+  metrics['scalar'].update({
+      'perf/period_step_count': num_step
+  })
+  metrics['scalar'].update({
+      'perf/step_time_seconds': step_time_delta.total_seconds() / num_step
   })
   metrics['scalar'].update({
       'perf/per_device_tflops' : per_device_tflops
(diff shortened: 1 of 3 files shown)


src/maxdiffusion/models/train.py
```diff
@@ -489,20 +493,18 @@ def compute_loss(unet_params):
...
-        last_step_completion = new_time
         step_num = step + 1
         samples_count = total_train_batch_size * step_num
+
+        if config.write_metrics and step % config.metrics_period == 0:
+            new_time = datetime.datetime.now()
+            step_time_delta = new_time - last_step_completion
+            record_scalar_metrics(train_metric, step_time_delta, config.metrics_period, per_device_tflops, learning_rate_scheduler(step))
+            write_metrics(writer, local_metrics_file, running_gcs_metrics, train_metric, step, config)
+            mllog_utils.maybe_train_step_log(config, start_step, step_num, samples_count, train_metric)
```
suexu1025: we can align `train_log_interval` with config. metrics_period

src/maxdiffusion/models/train.py
```diff
@@ -526,10 +528,12 @@ def compute_loss(unet_params):
         if step == last_profiling_step:
             max_utils.deactivate_profiler(config)
 
-        mllog_utils.maybe_train_step_log(config, start_step, step_num, samples_count, train_metric)
-
-    steptime = (time.time() - start_time)/ (config.max_train_steps - start_step) * 1000
-    max_logging.log(f"avg step time of {max_train_steps}, {steptime} ms")
+    if config.write_metrics:
+        # log the last metrics_period
```
suexu1025, line 545: we probably also need to update `record_scalar_metrics` here?

suexu1025 (changes requested):
Thanks for modifying the logs, just a few comment.

src/maxdiffusion/models/train.py
aireenmei: Thanks for catching this! Updated

src/maxdiffusion/models/train.py
aireenmei: write metrics is always one period behind record_scalar_metrics. In the training loop, we call both record_scalar_metrics and write_metrics to have current period metrics recorded, but print the previous period metrics. when the training loop finishes, all the periods have been recorded, so we only need to print the late period metrics. I updated and applied the same logic to mllog_utils.maybe_train_step_log

suexu1025 (approved):
Thanks!

PR #84, opened 2024-06-26
improve logging and add flag

aireenmei:


src/maxdiffusion/configs/base_2_base.yml
```diff
@@ -19,6 +19,7 @@ metrics_file: "" # for testing, local file that stores scalar metrics. If empty,
 write_metrics: True
 gcs_metrics: False
 metrics_period: 100  # calculate step time and loss
+metrics_buffer: false
```
suexu1025: @aireenmei can we resue write_metrics flag? if write_metrics set, we will do ping-pong metrics out. if not, no metrics file will be generated.

suexu1025 (commented):
thanks, just a few comment.

src/maxdiffusion/configs/base_2_base.yml
aireenmei: I currently use write_metrics=false to turn off any step time logging (including mllog) to get a clean baseline, that help access any logging overhead. See my comments https://buganizer.corp.google.com/issues/343225248#comment6

suexu1025: we already got two flags for metrics write out, `write_metrics`, `gcs_metrics`, we can apply buffer metrics as default for metrics write out. might not need to introduce new flag

src/maxdiffusion/models/train.py
```diff
@@ -491,32 +491,43 @@ def compute_loss(unet_params):
...
-            _buffered_sample_count = total_train_batch_size * _buffered_step_num
+            step_num_delta = step_num - last_log_step_num
+
             # record metrics of current period
             record_scalar_metrics(train_metric, step_time_delta, step_num_delta, per_device_tflops, learning_rate_scheduler(step))
-            # print metrics of previous period
-            mllog_utils.maybe_train_step_log(config, start_step, _buffered_step_num, _buffered_sample_count, _buffered_metrics)
-            write_metrics(writer, local_metrics_file, running_gcs_metrics, train_metric, step, config)
+
+            if config.metrics_buffer:
```
suexu1025: while mllog for this branch mlperf  branch will always turned on. we can use write_metric here to make write metrics out as optional.

aireenmei: Now leaving only one flag, write_metrics for using buffer. The default behavior is no buffer with mllog on

src/maxdiffusion/configs/base_2_base.yml
aireenmei: done

suexu1025 (approved):
Thanks, it is really helpful!

aireenmei:
This run is with the updates: https://cloudlogging.app.goo.gl/MHkDvvhTpFrjmvjy8

PR #90, opened 2024-07-16
Dreambooth training

entrpn:
- Adds dreambooth training
- Refactors unit tests with new config settings.
- Load sdxl checkpoints from gcs to save disk space.


src/maxdiffusion/maxdiffusion_utils.py
```diff
@@ -162,21 +172,72 @@ def get_dummy_unet_inputs(config, pipeline):
     time_ids_channels = pipeline.unet.projection_class_embeddings_input_dim - text_embeds_dim
     time_ids_dims = time_ids_channels // pipeline.unet.addition_time_embed_dim
     added_cond_kwargs = {
-      "text_embeds": jnp.zeros((batch_size, text_embeds_dim), dtype=dtype),
-      "time_ids": jnp.zeros((batch_size, time_ids_dims), dtype=dtype),
+      "text_embeds": jnp.zeros((batch_size, text_embeds_dim), dtype=config.weights_dtype),
```
suexu1025, line 175: should all those in activation_dtype instead of weight type  or extra input dtype?

entrpn: @suexu1025 I kept them as the same type as weights for simplicity and not to add another dtype in the config. Let me know what you think.

suexu1025: How about we just pass it as fp32, as we only provide fp32 input data to be less error prone?

suexu1025 (commented):
Thanks, others look good!

suexu1025 (approved):

PR #105, opened 2024-09-19
Refactoring Docker Build Process

parambole:
- Build Process now support different MODES={nightly,stable,stable_stack}
- Build Process is along the same lines as Maxtext which makes it easier to work with XPK


entrpn:
@parambole can you add some instructions on how to use these files and how to get set up with xpk in the README?

parambole:
> @parambole can you add some instructions on how to use these files and how to get set up with xpk in the README?

@entrpn As discussed this requires a larger refactoring of the README. I am planning to pick it up as part of https://github.com/google/maxdiffusion/issues/88. 

PLMK if that sounds good?

entrpn (approved):

Issue #88, opened 2024-07-09
Add instructions in the readme on how to get started with xpk 


parambole:
Ideally we want to swap out this [section](https://github.com/google/maxdiffusion?tab=readme-ov-file#getting-started-multihost-development).

PR #106, opened 2024-09-23
Refactoring Maxdiffusion readme

parambole:
https://github.com/AI-Hypercomputer/maxdiffusion/issues/88
- Updated the documentation structure
- Added readme for running it through XPK


docs/getting_started/run_maxdiffusion_via_xpk.md
entrpn, line 1: did you test that sdxl training works with v5e using DDP? I believe it only works with FSDP.

parambole: @entrpn No, I wanted to give a simple example command, and it looks like the command and the tpu-type wouldn't work. I have updated it with a simpler command to convey what is required in the README.md.

entrpn (approved):

PR #108, opened 2024-09-24
fix guidance rescale.

entrpn:
Fixes guidance rescale in sdxl where it was erroring out due to traced boolean value.

### DIFF
--- a/src/maxdiffusion/generate.py
+++ b/src/maxdiffusion/generate.py
@@ -71,6 +71,8 @@ def loop_body(step, args, model, pipeline, prompt_embeds, guidance_scale, guidan
     noise_pred = noise_pred_uncond + guidance_scale * (noise_prediction_text - noise_pred_uncond)
 
     # Based on 3.4. in https://arxiv.org/pdf/2305.08891.pdf
+    # Helps solve overexposure problem when terminal SNR approaches zero.
+    # Empirical values recomended from the paper are guidance_scale=7.5 and guidance_rescale=0.7 
     noise_pred = jax.lax.cond(
       guidance_rescale[0] > 0,
       lambda _: rescale_noise_cfg(noise_pred, noise_prediction_text, guidance_rescale),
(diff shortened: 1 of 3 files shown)


src/maxdiffusion/generate.py
```diff
@@ -71,7 +71,12 @@ def loop_body(step, args, model, pipeline, prompt_embeds, guidance_scale, guidan
     noise_pred = noise_pred_uncond + guidance_scale * (noise_prediction_text - noise_pred_uncond)
 
     # Based on 3.4. in https://arxiv.org/pdf/2305.08891.pdf
-    noise_pred = rescale_noise_cfg(noise_pred, noise_prediction_text, guidance_rescale=guidance_rescale)
```
parambole, line 74: nit: 
Adding a comment around what is being done "This technique aims to improve image quality by adjusting the noise prediction."

src/maxdiffusion/generate_sdxl.py
```diff
@@ -68,15 +68,20 @@ def loop_body(step, args, model, pipeline, added_cond_kwargs, prompt_embeds, gui
...
   if config.do_classifier_free_guidance:
-    noise_pred = apply_classifier_free_guidance(noise_pred, guidance_scale, guidance_rescale)
+    noise_pred, noise_prediction_text = apply_classifier_free_guidance(noise_pred, guidance_scale)
+
+  noise_pred = jax.lax.cond(
+      guidance_rescale[0] > 0,
+      lambda _: rescale_noise_cfg(noise_pred, noise_prediction_text, guidance_rescale),
+      lambda _: noise_pred,
+      operand=None
+  )
```
parambole, line 87: nit: Since this call is being used in both of the files is there a way to put it in a common utils file ?

entrpn: I don't think the jax conditional needs to be refactored. The main code is in the rescale_noise_cfg fn.

src/maxdiffusion/generate.py
entrpn: Added comments to explain what this does.

parambole (approved):

PR #119, opened 2024-10-03
Fixing incorrect link redirect

parambole:
Github redirects the link to folder instead of resolving it with `https://`


ssusie (approved):
Thanks

PR #121, opened 2024-10-08
Support building GPU docker image for MaxDiffusion Model

sizhit2:
Add device option in `docker_build_dependency_image.sh`. The default option is still Tpu. To build docker image with GPU, just specify `bash docker_build_dependency_image.sh DEVICE=gpu` 
Add `maxdiffusion_gpu_dependencies.Dockerfile` .


maxdiffusion_gpu_dependencies.Dockerfile
```diff
@@ -0,0 +1,49 @@
+# syntax=docker/dockerfile:experimental
+ARG BASEIMAGE=ghcr.io/nvidia/jax:base
```
parambole: @sizhit2 

**Nit:**

We can probaly add a link from where the base image is being dervice or a description even though navigating using the base image link would be sufficient ) i.e. why is this the default
Also probably a Note that this pulls in the latest of the jax:base i.e. built nightly AFAIR and that the tag changes so only the tag ( dated image on that day is stable i.e. does not chage )

**Super Nit:**

If you feel like contributing back to Jax tool box build process you could add a maxdiffusion test since it now depends on jax toolbox base as one of the CI/CD [test](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/container/test-maxtext.sh).

parambole (approved):
LGTM. Thanks for adding this functionality. 

**Nit:** Can you please update the README with the instructions?

docker_build_dependency_image.sh
```diff
@@ -46,6 +46,11 @@ if [[ -z ${MODE} ]]; then
   echo "Default MODE=${MODE}"
 fi
 
+if [[ -z ${DEVICE} ]]; then
+  export DEVICE=tpu
+  echo "Default DEVICE=${DEVICE}"
```
wang2yn84, line 51: Can we move this echo outside of the if statement so that it always prints the default device type?

docker_build_dependency_image.sh
```diff
@@ -55,22 +60,26 @@ COMMIT_HASH=$(git rev-parse --short HEAD)
 
 echo "Building MaxDiffusion with MODE=${MODE} at commit hash ${COMMIT_HASH} . . ."
 
-if [[ "${MODE}" == "stable_stack" ]]; then
-  if [[ ! -v BASEIMAGE ]]; then
-    echo "Erroring out because BASEIMAGE is unset, please set it!"
-    exit 1
+if [[ ${DEVICE} == "gpu" ]]; then
```
wang2yn84, line 64: Can you added a "pinned" mode here?

wang2yn84 (commented):
Thank you for adding the GPU support! Left couple of comments.

docker_build_dependency_image.sh
sizhit2: Wouldn't it be confusing if we set `DEVICE=gpu`, but we are echoing `Default DEVICE=tpu`? Or do you suggest we always echo the device information we are currently using?

wang2yn84: We won't get the device as TPU if we set to GPU. This piece of code only sets the device type as TPU is no device type is assigned. So if you assign device type as GPU, you will get GPU.

sizhit2: I see what you are suggesting. I add additional line to always print current DEVICE.

docker_build_dependency_image.sh
sizhit2: Done.

setup.sh
```diff
@@ -42,26 +46,61 @@ if [[ -n $JAX_VERSION && ! ($MODE == "stable" || -z $MODE) ]]; then
...
 # Install JAX and JAXlib based on the specified mode
-if [[ "$MODE" == "stable" || ! -v MODE ]]; then
+if [[ "$MODE" == "pinned" ]]; then
+  if [[ "$DEVICE" != "gpu" ]]; then
+    echo "pinned mode is supported for GPU builds only."
+    exit 1
+  fi
+  echo "Installing pinned jax, jaxlib for NVIDIA gpu."
+  pip3 install "jax[cuda12_pip]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html -c constraints_gpu.txt
+  pip3 install "transformer-engine==1.5.0+297459b" \
```
wang2yn84: We don't need transoformer-engine for Maxdiffusion. This additional deps slows down the initial process a lot.

constraints_gpu.txt
wang2yn84, line 1: We don't need this file.

wang2yn84, line 1: Consider update the title of this PR as: Support building GPU docker image for MaxDiffusion Model.

setup.sh
```diff
@@ -15,11 +15,10 @@
 # limitations under the License.
 
 # Description:
-# bash setup.sh MODE={stable,nightly}
+# bash setup.sh MODE={stable,nightly,pinned} DEVICE={tpu,gpu}
```
parambole: @sizhit2 Can we add a comment stating that not all combinations are supported, e.g., TPU and pinned are not supported?

src/maxdiffusion/max_utils.py
```diff
@@ -150,27 +150,6 @@ def add_text_to_summary_writer(key, value, summary_writer):
     summary_writer.add_text(key, value)
 
 
-def write_metrics_for_gcs(metrics, step, config, running_metrics):
```
parambole, line 153: @sizhit2 Why are we removing this ?

gpu_multi_process_run.sh
```diff
@@ -0,0 +1,156 @@
...
+      export NCCL_FASTRAK_USE_SNAP=1
+      export NCCL_MIN_NCHANNELS=4
+      export NCCL_SHIMNET_GUEST_CONFIG_CHECKER_CONFIG_FILE=/usr/local/nvidia/lib64/a3plus_guest_config.textproto
+      export NCCL_TUNER_CONFIG_PATH=/usr/local/nvidia/lib64/a3plus_tuner_config.textproto
+      export NCCL_TUNER_PLUGIN=libnccl-tuner.so
+    fi
+  else
+    echo "NOT using GPUDirect"
+  fi
+}
```
parambole, line 78: @sizhit2 Currently are we deploying the workload using XPK ? I am trying to understand which components determines the source of through for these flags.

src/maxdiffusion/max_utils.py
sizhit2: Because there is a function with same function signature defined in line 120. I don't think we will ever call this one...

gpu_multi_process_run.sh
wang2yn84: @parambole , yes we are deploying through xpk.  It will execute this file and execute this function. Can you elaborate on "which components determines the source of through for these flags"? What does it mean?

setup.sh
sizhit2: Sorry for the confusion. We don't support GPU pinned mode at this moment. Let me delete the pinned mode here.

gpu_multi_process_run.sh
parambole: @wang2yn84 Oh I see in the past there was a conversation around XPK setting the required env variables. But as you mentioned I still see XPK calling this file for that. 

ref: https://github.com/AI-Hypercomputer/xpk/blob/d88b092dc71a9d3f6d06fc8984370de0fbcbbe51/src/xpk/core/core.py#L2133


> "which components determines the source of through for these flags"?

Here the two components are XPK and the workload I thought XPK was taking care of setting the envvar and all the other setup related to running a workload on A3 series like installing the NCCL plugin and running the side-car container.

parambole (approved):

gpu_multi_process_run.sh
wang2yn84: I see. Do you have a pointer to that? What's the conclusion?

parambole: > I see. Do you have a pointer to that? What's the conclusion?

I am not sure what the conclusion was. Probably @gobbleturk might have more context.

PR #127, opened 2024-10-24
Hyper SDXL Lora support

entrpn:
This PR establishes LoRA support and includes Hyper-SD XL LoRA loading for inference.


src/maxdiffusion/loaders/lora_base.py
```diff
@@ -0,0 +1,106 @@
...
+      use_auth_token,
+      revision,
+      subfolder,
+      user_agent,
+      allow_pickle,
+  ):
+    from .lora_pipeline import LORA_WEIGHT_NAME_SAFE
+
+    model_file = None
+    if not isinstance(pretrained_model_name_or_path_or_dict, dict):
```
anfals, line 50: `pretrained_model_name_or_path_or_dict` feels like a very overloaded parameter now. Maybe instead this method could take in `state_dict` as an argument and just have `pretrained_model_name_or_path`?

src/maxdiffusion/loaders/lora_pipeline.py
```diff
@@ -0,0 +1,280 @@
...
+        if module_path in unet_lora_keys:
+          lora_layer = cls._get_lora_layer(module_path, context.module, rank, network_alphas)
+          return lora_layer(h, *args, **kwargs)
+      return h
+
+    return _intercept
+
+  @classmethod
+  @validate_hf_hub_args
+  def lora_state_dict(cls, pretrained_model_name_or_path: str, **kwargs):
```
anfals, line 158: nit: I think calling this something like `get_lora_state_dict` might be clearer about what the method is doing?

src/maxdiffusion/loaders/lora_pipeline.py
```diff
@@ -0,0 +1,280 @@
...
+
+  @classmethod
+  def make_lora_interceptor(cls, params, rank, network_alphas):
+    # Only unet interceptor supported for now.
+    unet_lora_keys = flax.traverse_util.flatten_dict(params["unet"]).keys()
+    unet_lora_keys, network_alphas = cls.rename_for_interceptor(unet_lora_keys, network_alphas)
+
+    def _intercept(next_fn, args, kwargs, context):
+      mod = context.module
+      while mod is not None:
```
anfals, line 142: Can you explain what's happening with this while loop here?

src/maxdiffusion/loaders/lora_conversion_utils.py
```diff
@@ -0,0 +1,610 @@
+# Copyright 2024 The HuggingFace Team. All rights reserved.
```
anfals, line 1: HuggingFace copyright?

anfals (commented):
Review in progress

Still need to look at src/maxdiffusion/models/modeling_flax_pytorch_utils.py and src/maxdiffusion/loaders/lora_conversion_utils.py

src/maxdiffusion/loaders/lora_base.py
entrpn: This file is mostly a copy from diffusers as maxdiffusion tries to stay compatible with loading weights from HF hub. Here's the original file : https://github.com/huggingface/diffusers/blob/main/src/diffusers/loaders/lora_base.py

src/maxdiffusion/loaders/lora_pipeline.py
entrpn: This is to prevent recursion. Because LoRA layers are being added dynamically, it can cause an endless loop as they can be "re-intercepted".

src/maxdiffusion/loaders/lora_conversion_utils.py
entrpn: maxdiffusion uses files from diffusers, this one being one of them. Per license, the copyright is kept with the original author.

src/maxdiffusion/loaders/lora_base.py
anfals: Ah thanks for this context! Bit off an odd choice on HF's part IMO, but I see that this is used all over, so not worth the effort to correct.

src/maxdiffusion/loaders/lora_pipeline.py
anfals: Ah. For my understanding of how these interceptors work, say originally we had something like `Linear -> Non-linearity -> Linear`. Say we get a LoRa that's added in dynamically, so we now have  `Linear -> LoRa -> Non-linearity -> Linear`. Is this trying to prevent the case where the LoRa layer gets intercepted and another LoRa layer gets added? Does the check on Line 149 not catch this?

src/maxdiffusion/models/modeling_flax_pytorch_utils.py
anfals, line 1: I'm mostly just skimming this file since it seems to just be a lot of wrangling of PyTorch state_dict to Flax. Someone who works more on MaxDiffusion might have more thoughts

src/maxdiffusion/loaders/lora_conversion_utils.py
anfals: gotcha; i'm guessing this file is mostly just copy/paste then, so I'm not gonna read through all of it - but let me know if there's something in particular you wanted more eyes on :)

anfals (dismissed):
Great work! Approving, but I did have a question on the infinite loop avoidance logic.

src/maxdiffusion/loaders/lora_pipeline.py
entrpn: Yes that is correct. Line 149 would catch this, but I added this for completeness, future proofing.

requirements.txt
```diff
@@ -26,4 +26,6 @@ git+https://github.com/mlperf/logging.git
 opencv-python-headless==4.10.0.84
 orbax-checkpoint>=0.5.20
 tokenizers==0.20.0
+huggingface_hub==0.24.7
```
anfals, line 29: Repeated line?

src/maxdiffusion/loaders/lora_pipeline.py
anfals: Got it - it might be useful to add a comment explaining what's happening here, but that can be done in a follow-up PR.

anfals (approved):

PR #133, opened 2024-11-12
support grain checkpoint

aireenmei:
Grain data input pipeline is already supported in this previous PR: https://github.com/AI-Hypercomputer/maxdiffusion/pull/130
This PR added ckpt support, so that there is an additional "iter" folder in the ckpt for data iterator to recover the data order resume from the last used index.


parambole:
@aireenmei 

Nit: Could you add more details to the description ?

parambole (approved):

PR #141, opened 2025-01-27
WIP flax.linen flux.1 ported from nnx jflux

jcaraban:
Background in https://github.com/AI-Hypercomputer/maxdiffusion/issues/115

Note @entrpn this PR is not in a mergeable state, but it runs so you can see where we stand. We are working on integrating FA and multi-node (from MaxText) so we are bound to mess the code even more. Hence it might be easier if you cherry-pick our changes rather than merging what we have, but let's discuss that.

TODO:
- [ ] rename JFlux to FlaxFlux or whatever appropriate
- [ ] clean max_utils.py, we copied logic from MaxText to test multi-node
- [ ] check checkpointing logic, in case we applied some dubious fixes
- [ ] double-check Flux TFlop calculation, might be 10% off
- [ ] migrate ae_flux_nnx.py and HFEmbedder(nnx.Module) to linen?
- [ ] how to get rid of autoencoder model torch->jax porting?
- [ ] verify and improve sharding
- [ ] verify the time scheduler

PR #146, opened 2025-02-05
Flux inference implementation

entrpn:
This PR implements Flux dev and schnell for inference.

Features:
- Loads encoders, vae and transformer to HBM with encoder offload option. All settings available in the flux configs.
- Optimized flash block sizes for v6e.
- Flash attention
- Sharding - DDP and FSDP, allows loading all models on HBM and keeping them loaded using FSDP sharding, while reducing performance slightly. 
- Removed maxdiffusion.transformers and move back to HF transformers library due to support of new models. Originally I ported some of the models from transformers into maxdiffusion.transformers but saw no improvements in sharding them. This will reduce maintenance of those models inside of maxdiffusion.


jiangjy1982 (approved):

jcaraban:
@entrpn I will close https://github.com/AI-Hypercomputer/maxdiffusion/pull/141 since this PR is in a mergable state now
@ksikiric rebased our training on top of [flux_impl](https://github.com/AI-Hypercomputer/maxdiffusion/tree/flux_impl) and opened a continuation PR https://github.com/AI-Hypercomputer/maxdiffusion/pull/147
After that we still plan to submit another PR with flash_attn support for GPUs, alike to MaxText

PR #147, opened 2025-02-12
Training added on top of flux_impl

ksikiric:
Linked to #146

I've added the training code on top of https://github.com/AI-Hypercomputer/maxdiffusion/tree/flux_impl. This PR is meant to be merged after https://github.com/AI-Hypercomputer/maxdiffusion/pull/146. 

With the training code, I have also added a pipeline for flux, which can be used for inference as well.

PR #148, opened 2025-02-13
Flux lora

entrpn:
- Adds support for loading Flux LoRAs for inference.


entrpn:
@jcaraban @ksikiric this is the PR for loading LoRA with Flux. Is it possible to rebase the [Flux training PR](https://github.com/AI-Hypercomputer/maxdiffusion/pull/147) and continue working from there?

src/maxdiffusion/generate_flux.py
```diff
@@ -39,6 +41,28 @@
...
+      params, rank, network_alphas = lora_loader.load_lora_weights(
+          config,
+          lora_config["lora_model_name_or_path"][i],
+          weight_name=lora_config["weight_name"][i],
+          params=params,
+          adapter_name=lora_config["adapter_name"][i],
+      )
+      interceptor = lora_loader.make_lora_interceptor(params, rank, network_alphas, lora_config["adapter_name"][i])
+      interceptors.append(interceptor)
+  return params, interceptors
```
bvandermoon, line 65: Nit: Why not use [maybe_load_lora()](https://github.com/AI-Hypercomputer/maxdiffusion/blob/7f0f5bc928174cce5880a90917f52d181fe02227/src/maxdiffusion/maxdiffusion_utils.py#L42) from maxdiffusion_utils?

bvandermoon (approved):
LGTM

src/maxdiffusion/generate_flux.py
entrpn: maybe_use_lora() in maxdiffusion_utils is specific to sdxl and won't work with flux. I should rename that method to maybe_load_sdxl_lora. I will create an issue to track this and add it on a different commit. Thanks for the review.

PR #152, opened 2025-03-01
Remove the transformer engine from setup script 

wang2yn84:
The TransformerEngine is not required and it slows down the setup a lot. Removed and updated the read me. Also added error message if the installation fails for the last 2 steps.


entrpn:
@wang2yn84 I was thinking we should integrate TransformerEngine into the sdxl attention mechanism. It is currently used for Flux and provides a speedup in Flux, so it might do the same for sdxl.

wang2yn84:
Hi Juan, I see. Let's close the PR.

PR #187, opened 2025-06-18
Bug fixes and performance optimizations for FLUX training

hx89:
Bug fixes:
1. Fix the text sequence length configuration issue
2. Fix the race condition when clean up the encoder cache in multi-process run

Performance optimizations:
1. Fix the FSDP sharding in fc2
2. Remove sharding of the bias to avoid all-gather
3. Remove the sharding of rope embedding to avoid all-gather
4. Fix the sharding of time step embedding to avoid all-to-all

Also added nsys profiler support

### DIFF
--- a/src/maxdiffusion/max_utils.py
+++ b/src/maxdiffusion/max_utils.py
@@ -26,7 +26,6 @@ import yaml
 import os
 from pathlib import Path
 import subprocess
-from ctypes import cdll
 import numpy as np
 
 import flax
@@ -63,8 +62,6 @@ from tensorboardX import writer
 
 from google.cloud import storage
 
-libcudart = cdll.LoadLibrary("libcudart.so")
-
 FrozenDict = core.frozen_dict.FrozenDict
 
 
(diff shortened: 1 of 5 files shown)


coolkp:
Can you git rebase instead of merge to preserve the commit history?

src/maxdiffusion/max_utils.py
```diff
@@ -63,6 +63,8 @@
 
 from google.cloud import storage
 
+libcudart = cdll.LoadLibrary("libcudart.so")
```
coolkp: This is breaking the pytest, the pytest workflow right now only supports TPU. The dependencies are not installed. https://github.com/AI-Hypercomputer/maxdiffusion/blob/eaef265428ef200f8260e4edb017f0b59e839f1b/.github/workflows/UnitTests.yml

hx89: I have removed nsys and rebased to main.

coolkp (approved):
Thanks for the contribution!

PR #199, opened 2025-07-02
generates ssim score between pretrained and trained model.

entrpn:


src/maxdiffusion/trainers/wan_trainer.py
```diff
@@ -31,13 +31,32 @@
 from maxdiffusion.input_pipeline.input_pipeline_interface import (make_data_iterator)
 from maxdiffusion.generate_wan import run as generate_wan
 from maxdiffusion.train_utils import (_tensorboard_writer_worker, load_next_batch, _metrics_queue)
+from ..video_processor import VideoProcessor
```
coolkp: nit: Can we use maxdiffusion.video_processor and maxdiffusion.utils for the imports

src/maxdiffusion/trainers/wan_trainer.py
```diff
@@ -31,13 +31,32 @@
...
+from ..utils import load_video
+from skimage.metrics import structural_similarity as ssim
```
coolkp, line 36: nit: we should follow up, pin the scikit-image dependency

coolkp (dismissed):

coolkp (approved):

PR #200, opened 2025-07-09
Add Context parallelism to Wan 2.1

entrpn:
* Adds context parallelism to flash attention fn.
* Adds better sharding constraints when reshaping and padding to reduce AGs.


src/maxdiffusion/max_utils.py
```diff
@@ -281,9 +281,9 @@ def create_device_mesh(config, devices=None, logging=True):
   ici_parallelism = fill_unspecified_mesh_axes(ici_parallelism, num_devices_per_slice, "ICI")
   if multi_slice_env:
     dcn_parallelism = fill_unspecified_mesh_axes(dcn_parallelism, num_slices, "DCN")
-    mesh = mesh_utils.create_hybrid_device_mesh(ici_parallelism, dcn_parallelism, devices)
+    mesh = mesh_utils.create_hybrid_device_mesh(ici_parallelism, dcn_parallelism, devices, allow_split_physical_axes=config.allow_split_physical_axes)
```
coolkp: I also discovered that cloud tpus, don't have "slice_index" attribute which is used in create_hybrid_mesh and also couple lines up to determine dcn or not, maybe we should pass process_is_granule=True to  create_hybrid_device_mesh
image

src/maxdiffusion/models/attention_flax.py
```diff
@@ -110,33 +112,43 @@ def _unflatten_heads(tensor, heads):
...
 
-  if kv_size < 128 or rem != 0:
+  final_padded_len = num_blocks * flash_block_size
+  seq_len_pad = final_padded_len - seq_len
+
+  if kv_size < 128 or seq_len_pad != 0:
     npad = ((0, 0), (0, 0), (0, seq_len_pad), (0, head_dim_pad))
-    tensor = jnp.pad(tensor, npad)
+    padded_tensor = jnp.pad(tensor, npad)
+    tensor = jax.lax.with_sharding_constraint(padded_tensor, PartitionSpec("data", "tensor", "fsdp", None))
```
coolkp, line 151: For my understanding purposes, this is needed because reshape is tricky op may not inherit sharding from input sequences where sequence dim was sharded over fsdp ranks. in first step to Attention block? But this is same so its a static layout (no comms).

src/maxdiffusion/models/attention_flax.py
```diff
@@ -663,6 +708,7 @@ def __init__(
         dtype=dtype,
         param_dtype=weights_dtype,
         precision=precision,
+        bias_init=nnx.with_partitioning(nnx.initializers.zeros, ("embed",)),
     )
 
     self.proj_attn = nnx.Linear(
```
coolkp, line 714: Here the weights are sharded (D/tensor,D/fsdp) and i think the attention output is sharded (D/fsdp,D/tensor) in default config we have it as tensor =1, fsdp=-1.  Won't the output from self attention be sharded both rowwise and colwise after self attention and it will be input to cross attention where we do a explicit sharding again `hidden_states = jax.lax.with_sharding_constraint(hidden_states, PartitionSpec("data", "fsdp", "tensor"))` So jax will insert an all gather over model_dim? Should we replicate the weights or is there memory concern? 
image

coolkp: in proj_attn i mean

src/maxdiffusion/models/attention_flax.py
coolkp: jax documentation 
> reshape. Reshape is a particularly tricky op. An output axis can map to more than one input axis (when reshape is used to merge axes) or just a part of an input axis (when reshape is used to split axes). Our usual rules don’t apply. Instead we treat reshape as follows. We strip away singleton axes (these can’t be sharded anyway. Then we decide whether the reshape is a “split” (splitting a single axis into two or more adjacent axes), a “merge” (merging two or more adjacent axes into a single one) or something else. If we have a split or merge case in which the split/merged axes are sharded as None then we shard the resulting split/merged axes as None and the other axes according to their corresponding input axis shardings. In all other cases we throw an error and require the user to provide an out_sharding argument.

entrpn: I added these sharding constraints based on xprof results of AGs hapenning in these lines, which are not present anymore after adding this code. I believe this helps guide the XLA compiler on better optimized layouts.

src/maxdiffusion/models/attention_flax.py
entrpn: I tested this, but it gave worse performance. There is a FP32 layer between cross and self attention, which gathers the activations, so the AG is there regardless. In the profile, this operation doesn't come up as a wasteful operation, so leaving it as is for now.

coolkp (approved):

PR #244, opened 2025-09-10
Update WAN to match new Flax version

susanbao:
Please help check whether change python requirement from 3.10 to 3.11 will cause error.


susanbao:
> @susanbao we will also need add installation python 3.11 in setup.sh?

Sure. Working on updating the setup.sh

PR #251, opened 2025-09-19
Cross self attention switch

entrpn:
- Overrides custom flash block sizes for cross attention to prevent extra padding on small cross attention sequence lengths.
- Same as https://github.com/AI-Hypercomputer/maxdiffusion/pull/250 but setting the sharding at the NNX module level instead of the splash attention kernel function.

### DIFF
--- a/tests/schedulers/test_scheduler_flax.py
+++ b/tests/schedulers/test_scheduler_flax.py
@@ -336,7 +336,7 @@ class FlaxDDPMSchedulerTest(FlaxSchedulerCommonTest):
 
         if jax_device == "tpu":
             assert abs(result_sum - 263.11) < 1.5e-2
-            assert abs(result_mean - 0.3349905) < 2e-5
+            assert abs(result_mean - 0.34259) < 2e-5
         else:
             assert abs(result_sum - 255.1113) < 1e-2
             assert abs(result_mean - 0.332176) < 1e-3


coolkp (dismissed):

entrpn:
Just realized that I also need to support ring with these new axis rules. If attention=ring, then force logical axis rules to change in pyconfig such that ring attention is supported.

src/maxdiffusion/configs/base_wan_14b.yml
```diff
@@ -145,15 +158,17 @@ mesh_axes: ['data', 'fsdp', 'tensor']
...
                       ['activation_length', 'fsdp'],
-                      
                       ['activation_heads', 'tensor'],
                       ['mlp','tensor'],
                       ['embed','fsdp'],
                       ['heads', 'tensor'],
                       ['norm', 'tensor'],
                       ['conv_batch', ['data','fsdp']],
                       ['out_channels', 'tensor'],
+                      ['conv_in', 'fsdp'],
```
entrpn: why is conv_in re-introduced?

src/maxdiffusion/pipelines/wan/wan_pipeline.py
```diff
@@ -585,7 +587,7 @@ def __call__(
             width=width,
             num_frames=num_frames,
             num_channels_latents=num_channel_latents,
-        )
+        ) # # fusion.18
```
entrpn: remove the fusion comment.

src/maxdiffusion/configs/base_wan_14b.yml
```diff
@@ -60,18 +60,20 @@ jit_initializers: True
 from_pt: True
 split_head_dim: True
 attention: 'flash' # Supported attention: dot_product, flash, cudnn_flash_te, ring
-flash_min_seq_length: 4096
+flash_min_seq_length: 0
+mask_padding_tokens: True # Whether to mask padding tokens in attention computation.
+attention_sharding_uniform: True # same sequence sharding rules applied for q in both (self and cross attention)
```
entrpn: can you clarify what this flag does.

coolkp: Added more clarification

src/maxdiffusion/pipelines/wan/wan_pipeline.py
coolkp: done

src/maxdiffusion/configs/base_wan_14b.yml
coolkp: removed

src/maxdiffusion/configs/base14.yml
```diff
@@ -50,6 +50,8 @@ jit_initializers: True
 from_pt: False
 split_head_dim: True
 attention: 'dot_product' # Supported attention: dot_product, flash
+mask_padding_tokens: True # Whether to mask padding tokens in attention computation.
```
entrpn: let's update the comments for these other configs for these flags as well and if its not relevant to them, please add a comment for this as well.

PR #254, opened 2025-09-26
New Eval Pipeline as the MLPerf

susanbao:
- Implement the new eval pipeline which matches that in MLPerf.

      ALGORITHM: Validation Loss Computation
      
      INPUT:
        - validation_samples: set of validation data samples
      
      INITIALIZE:
        - sum[8]: array of zeros for accumulating losses
        - count[8]: array of zeros for counting samples per timestep
      
      FOR each sample, timestep in validation_samples:
          loss = forward_pass(sample, timestep=t/8)
          sum[t] += loss
          count[t] += 1
      
      mean_per_timestep = sum / count
      validation_loss = mean(mean_per_timestep)
      
      RETURN validation_loss

- Add new para enable_ssim which help to debug.

### DIFF
--- a/src/maxdiffusion/configs/base_wan_14b.yml
+++ b/src/maxdiffusion/configs/base_wan_14b.yml
@@ -318,6 +318,6 @@ quantization_calibration_method: "absmax"
 eval_every: -1
 eval_data_dir: ""
 enable_generate_video_for_eval: False # This will increase the used TPU memory.
-eval_max_number_of_samples_in_bucket: 60
+eval_max_number_of_samples_in_bucket: 60 # The number of samples per bucket for evaluation. This is calculated by num_eval_samples / len(considered_timesteps_list).
 
 enable_ssim: True
--- a/src/maxdiffusion/configs/base_wan_14b.yml
+++ b/src/maxdiffusion/configs/base_wan_14b.yml
@@ -319,4 +319,3 @@ eval_every: -1
 eval_data_dir: ""
 enable_generate_video_for_eval: False # This will increase the used TPU memory.
 eval_max_number_of_samples_in_bucket: 60
-eval_max_processed_batch_size: 8 # This is the max batch size per device for eval step. If the global eval batch size is larger than this, the eval step will be run multiple times.
--- a/src/maxdiffusion/configs/base_wan_14b.yml
+++ b/src/maxdiffusion/configs/base_wan_14b.yml
@@ -319,3 +319,4 @@ eval_every: -1
 eval_data_dir: ""
 enable_generate_video_for_eval: False # This will increase the used TPU memory.
 eval_max_number_of_samples_in_bucket: 60
+eval_max_processed_batch_size: 8 # This is the max batch size per device for eval step. If the global eval batch size is larger than this, the eval step will be run multiple times.
--- a/src/maxdiffusion/configs/base_wan_14b.yml
+++ b/src/maxdiffusion/configs/base_wan_14b.yml
@@ -235,6 +235,8 @@ global_batch_size: 0
 tfrecords_dir: ''
 no_records_per_shard: 0
 enable_eval_timesteps: False
+considered_timesteps_list: [125, 250, 375, 500, 625, 750, 875]
+num_eval_samples: 420
 
 warmup_steps_fraction: 0.1
 learning_rate_schedule_steps: -1 # By default the length of the schedule is set to the number of steps.
@@ -316,3 +318,4 @@ quantization_calibration_method: "absmax"
 eval_every: -1
 eval_data_dir: ""
 enable_generate_video_for_eval: False # This will increase the used TPU memory.
+eval_max_number_of_samples_in_bucket: 60
--- a/src/maxdiffusion/configs/base_wan_14b.yml
+++ b/src/maxdiffusion/configs/base_wan_14b.yml
@@ -319,3 +319,5 @@ eval_every: -1
 eval_data_dir: ""
 enable_generate_video_for_eval: False # This will increase the used TPU memory.
 eval_max_number_of_samples_in_bucket: 60
+
+enable_ssim: True
--- a/src/maxdiffusion/configs/base_wan_14b.yml
+++ b/src/maxdiffusion/configs/base_wan_14b.yml
@@ -234,6 +234,7 @@ global_batch_size: 0
 # For creating tfrecords from dataset
 tfrecords_dir: ''
 no_records_per_shard: 0
+enable_eval_timesteps: False
 
 warmup_steps_fraction: 0.1
 learning_rate_schedule_steps: -1 # By default the length of the schedule is set to the number of steps.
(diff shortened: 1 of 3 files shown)


src/maxdiffusion/trainers/wan_trainer.py
```diff
@@ -305,24 +323,40 @@ def training_loop(self, pipeline, optimizer, learning_rate_scheduler, train_data
...
-          eval_rng = jax.random.key(self.config.seed + step)
-          eval_metrics = []
+          eval_rng = eval_rng_key
+          eval_losses_by_timestep = {}
           # Loop indefinitely until the iterator is exhausted
           while True:
             try:
               with mesh:
                 eval_batch = load_next_batch(eval_data_iterator, None, self.config)
+                eval_batch["timesteps"] = jax.device_put(
```
entrpn: why do we need to do a device_put for timesteps but not for other data?

src/maxdiffusion/configs/base_wan_14b.yml
```diff
@@ -315,3 +318,6 @@ quantization_calibration_method: "absmax"
 eval_every: -1
 eval_data_dir: ""
 enable_generate_video_for_eval: False # This will increase the used TPU memory.
+eval_max_number_of_samples_in_bucket: 60
```
coolkp: Can you write a comment describing what this parameter is for.

src/maxdiffusion/configs/base_wan_14b.yml
```diff
@@ -234,6 +234,9 @@ global_batch_size: 0
 # For creating tfrecords from dataset
 tfrecords_dir: ''
 no_records_per_shard: 0
+enable_eval_timesteps: False
```
coolkp, line 237: Can you group all evaluation related configurations together?

src/maxdiffusion/trainers/wan_trainer.py
```diff
@@ -394,57 +441,52 @@ def eval_step(state, data, rng, scheduler_state, scheduler, config):
...
+        deterministic=True,
     )
 
     # Calculate the loss against the target
     training_target = scheduler.training_target(latents, noise, timesteps)
     training_weight = jnp.expand_dims(scheduler.training_weight(scheduler_state, timesteps), axis=(1, 2, 3, 4))
     loss = (training_target - model_pred) ** 2
     loss = loss * training_weight
-    loss = jnp.mean(loss)
+    loss = loss.reshape(loss.shape[0], -1).mean(axis=1)
```
coolkp, line 469: Why do we need to transpose and mean and also not calculate mean over all axes. Does this change loss to vector? Is one of the dims time step dim?

src/maxdiffusion/trainers/wan_trainer.py
```diff
@@ -394,57 +441,52 @@ def eval_step(state, data, rng, scheduler_state, scheduler, config):
...
+  losses = jnp.zeros(bs)
+  for i in range(0, bs, single_batch_size):
+    start = i
+    end = min(i + single_batch_size, bs)
+    latents= data["latents"][start:end, :].astype(config.weights_dtype)
+    encoder_hidden_states = data["encoder_hidden_states"][start:end, :].astype(config.weights_dtype)
+    timesteps = data["timesteps"][start:end].astype("int64")
+    _, new_rng = jax.random.split(rng, num=2)
+    loss = loss_fn(state.params, latents, encoder_hidden_states, timesteps, new_rng)
+    losses = losses.at[start:end].set(loss)
```
coolkp, line 488: It would be cool to optimize this by vmapping this logic over batch dim. https://docs.jax.dev/en/latest/_autosummary/jax.vmap.html

src/maxdiffusion/trainers/wan_trainer.py
```diff
@@ -394,57 +441,52 @@ def eval_step(state, data, rng, scheduler_state, scheduler, config):
...
-
-    # Sample random timesteps and noise, just as in a training step
-    timesteps = jax.random.randint(
-        timestep_rng,
-        (bsz,),
-        0,
-        scheduler.config.num_train_timesteps,
-    )
-    noise = jax.random.normal(key=new_rng, shape=latents.shape, dtype=latents.dtype)
+    noise = jax.random.normal(key=rng, shape=latents.shape, dtype=latents.dtype)
```
coolkp, line 452: This value will be the same. if thats intended can we bring this out of the jit compiled function. Compiler will most likely fold it up as a constant and and it will increase compilation time

src/maxdiffusion/configs/base_wan_14b.yml
susanbao: These paras are only for the tfrecord evaluation dataset generation. I think it would be better to put them here.

src/maxdiffusion/trainers/wan_trainer.py
susanbao: Here we need to compute the loss for each sample so later we can compute the mean loss in each bucket. Yes, this change loss to vector. and the first dim is for batch size (timestep).

src/maxdiffusion/configs/base_wan_14b.yml
coolkp: Is eval time steps not supported for other evaluation datasets, it sounds to me like eval timesteps should be independent from data set type.

src/maxdiffusion/trainers/wan_trainer.py
coolkp: Thanks for the explanation , can you add a comment to explain this as well especially clarification about the dimension for sample and timestep.  Is the batch size and timestep dim same?  we are computing mean loss for each sample in each bucket? loss scalar's shape is number of samples, number of buckets?

src/maxdiffusion/trainers/wan_trainer.py
coolkp: Especially because we are splitting the key. https://stackoverflow.com/a/75339951m Feel free to leave a TODO for me

coolkp:
A tip for optimizing this PR. iiurc we have nested loop where we split RNG keys first in training loop and then in loop wrapping loss function. it would be good to generate the keys upfront and move the key out of the jit, iiurc eval_step is jitted function, so this will increase compile time too. https://stackoverflow.com/a/75339951 . We should vectorize that and move the key. you can give it a shot or create a TODO for me, i can take a look at it later.

src/maxdiffusion/trainers/wan_trainer.py
susanbao: Have leave the TODO

src/maxdiffusion/trainers/wan_trainer.py
susanbao: Number of losses = number of samples = number of timesteps. Each bucket corresponding to one timestep value and have 60 samples here.

src/maxdiffusion/configs/base_wan_14b.yml
susanbao: It can be used for generating other video dataset and not dataset dependence.

coolkp (approved):

PR #258, opened 2025-10-02
Fixes for FLUX

hx89:
Fix FLUX config and issues with the latest TE.

### DIFF
--- a/src/maxdiffusion/configs/base_flux_schnell.yml
+++ b/src/maxdiffusion/configs/base_flux_schnell.yml
@@ -236,6 +236,7 @@ enable_profiler: False
 # the iteration time a chance to stabilize.
 skip_first_n_steps_for_profiler: 5
 profiler_steps: 10
+profiler: ""
 
 # Generation parameters
 prompt: "A magical castle in the middle of a forest, artistic drawing"
@@ -284,3 +285,5 @@ quantization: ''
 quantization_local_shard_count: -1
 use_qwix_quantization: False 
 compile_topology_num_slices: -1 # Number of target slices, set to a positive integer.
+
+save_final_checkpoint: False
--- a/src/maxdiffusion/train_flux.py
+++ b/src/maxdiffusion/train_flux.py
@@ -42,10 +42,10 @@ def main(argv: Sequence[str]) -> None:
 
 @contextmanager
 def transformer_engine_context():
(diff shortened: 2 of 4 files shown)


src/maxdiffusion/train_flux.py
```diff
@@ -39,6 +40,24 @@ def main(argv: Sequence[str]) -> None:
   max_logging.log(f"Found {jax.device_count()} devices.")
   train(config)
 
+@contextmanager
+def transformer_engine_context():
+  """ If TransformerEngine is available, this context manager will provide the library with MaxText-specific details needed for correcct operation. """
+  try:
+    from transformer_engine.jax.sharding import global_shard_guard, MeshResource
+    # Inform TransformerEngine of MaxText's physical mesh resources.
```
coolkp: nit: maxdiffusion

src/maxdiffusion/configs/base_flux_schnell.yml
```diff
@@ -284,3 +285,5 @@ quantization: ''
 quantization_local_shard_count: -1
 use_qwix_quantization: False 
 compile_topology_num_slices: -1 # Number of target slices, set to a positive integer.
+
+save_final_checkpoint: False
```
coolkp, line 289: Why change default?

src/maxdiffusion/train_flux.py
```diff
@@ -39,6 +40,24 @@ def main(argv: Sequence[str]) -> None:
...
+      cp_resource = None,
+    )
+    with global_shard_guard(mesh_resource):
+      yield
+  except ImportError:
+    yield
 
 if __name__ == "__main__":
-  app.run(main)
+  with transformer_engine_context():
```
coolkp, line 44: Is this needed for sdxl too?

src/maxdiffusion/configs/base_flux_schnell.yml
hx89: Seems save_final_checkpoint is used in the code and will error out if the config yml file doesn't have it.

It's set to False in base_flux_dev.yml, so we are doing the same here.

src/maxdiffusion/train_flux.py
hx89: Yes it's needed for any model that uses TE to let TE know the physical mesh axes.

coolkp: Then can you add in the train_sdxl as well?

coolkp (approved):

PR #262, opened 2025-10-07
fix qwix unit test

susanbao:


src/maxdiffusion/tests/wan_transformer_test.py
```diff
@@ -65,16 +68,39 @@ def test_nnx_pixart_alpha_text_projection(self):
...
+    pyconfig.initialize(
+        [
+            None,
+            os.path.join(THIS_DIR, "..", "configs", "base_wan_14b.yml"),
+        ],
+        unittest=True,
+    )
+    config = pyconfig.config
+    devices_array = create_device_mesh(config)
+    mesh = Mesh(devices_array, config.mesh_axes)
```
coolkp: Do you mind moving this into the setup i think other tests will also need this

susanbao: NNXPixArtAlphaTextProjection and NNXTimestepEmbedding use axis_names=('embed', 'mlp') which is different from other tests.

PR #382, opened 2026-04-17
[LTX2] resolve flash attention block size mismatch and missing config overrides

mbohlool:
What does this PR do?
This PR fixes a crash occurring in LTX-2 when using specific frame counts (e.g., num_frames=121) with Flash Attention, and fixes a pipeline bug that prevented users from manually overriding the cross-attention kernels.

The Root Causes & Fixes:
1. Ignored Kernel Config Overrides (ltx2_pipeline.py)

Bug: Passing a2v_attention_kernel=dot_product via CLI or YAML had no effect. The pipeline was only mapping the main attention config, dropping the cross-attention kernel parameters before initializing the transformer.

Fix: Added mapping for a2v_attention_kernel and v2a_attention_kernel into ltx2_config inside create_sharded_logical_transformer so user overrides are respected.

2. Flash Attention Block Size Mismatch (attention_flax.py)

Bug: Generating 121 frames results in 126 audio latent tokens. A previous PR correctly padded this sequence from 126 to 128 to satisfy shard_map context chunking requirements. However, _tpu_flash_attention was calling _select_flash_block_sizes before the padding occurred. Because of the min() bounds used for cross-attention optimization, the block size was calculated as 126. This resulted in passing a padded sequence of 128 to the Splash Attention kernel but telling it to use a block size of 126, crashing because 128 % 126 != 0.

Fix: Swapped the order of operations in _tpu_flash_attention. Sequences are now padded by _reshape_data_for_flash before block sizes are calculated. This ensures _select_flash_block_sizes sees the padded shape, correctly calculating a divisible block size without removing the memory optimizations needed for the cross-attention unit tests to pass.

Testing
Verified that running generation with num_frames=121 executes cleanly on TPU with Flash Attention enabled.

Verified pytest src/maxdiffusion/tests/attention_test.py passes.

### DIFF
--- a/src/maxdiffusion/models/attention_flax.py
+++ b/src/maxdiffusion/models/attention_flax.py
@@ -287,11 +287,11 @@ def _tpu_flash_attention(
 ) -> jax.Array:
   """TPU Flash Attention"""
 
-  block_sizes = _select_flash_block_sizes(query, key, flash_block_sizes, dtype, attention_kernel)
   num_context_shards = mesh.shape["context"]
   query, orig_q_seq_len = _reshape_data_for_flash(query, heads, num_context_shards)
   key, _ = _reshape_data_for_flash(key, heads, num_context_shards)
   value, _ = _reshape_data_for_flash(value, heads, num_context_shards)
+  block_sizes = _select_flash_block_sizes(query, key, flash_block_sizes, dtype, attention_kernel)
 
   q_axis_names = nn.logical_to_mesh_axes(axis_names_q)
   kv_axis_names = nn.logical_to_mesh_axes(axis_names_kv)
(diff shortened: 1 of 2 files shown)


src/maxdiffusion/models/attention_flax.py
```diff
@@ -222,13 +222,13 @@ def _select_flash_block_sizes(
   block_size_q = flash_block_sizes.block_q if flash_block_sizes else q_max_block_size
   return splash_attention_kernel.BlockSizes(
       block_q=block_size_q,
-      block_kv_compute=min(kv_max_block_size, key_seq_len),
-      block_kv=min(kv_max_block_size, key_seq_len),
+      block_kv_compute=kv_max_block_size,
```
prishajain1, line 225: the min() earlier was serving as a safe guard against using a block size which might be larger than seq length for cross attention case. Suggest to keep it. Instead we can move `_select_flash_block_sizes` after  calls to `_reshape_data_for_flash` in tpu_flash_attention

Perseus14 (approved):

entrpn (approved):

PR #385, opened 2026-04-19
LTX2 Performance tuning

prishajain1:
This PR adds features which lead to performance gains for LTX-2 model, along with a fix for the broken LTX-2 Upsampler in main

### Performance enhancement features:
Experiments performed on v7x-8

- Sharding fix in `NNXSImpleFeedForward` : Data is sharded along the sequence dimension, each device holds a subset of tokens, but full feature channels. Because the input data had replicated features but the weights expected sharded features on the same physical axis (context), XLA was forced to insert an All-Gather on the sequence dimension to resolve the layout conflict, resulting in high wasted time. With our fix:
     - Overall % of wasted time in all-gathers went from 52.56% to 38.07%
     - Generation time per video dropped from 20s to 16.7s

- QKV Projection Sharding fix (ironwood specific): The profiling showed that the input data was being All-Gathered along the sequence dimension triggered by the QKV Projection step. Because the weights were sharded on the dimension that needed to be summed over (features), a single device could not complete the matrix multiplication using only its local shard of the data. To resolve this, XLA automatically inserted an All-Gather to replicate the sequence dimension across all devices before performing the multiplication. We changed the weight sharding in `attention_ltx2.py` to remove sharding on the input feature dimension.
     - Overall % of wasted time in all-gathers went from 38.07% to 19.39%
     - Generation time per video dropped from 16.7 to 13.84s
   
- Batching in text encoder: With CFG enabled, we see two passes of text encoder: one each for positive and negative prompts. If Classifier-Free Guidance is enabled, we concatenate the positive prompt and negative prompt and instead of doing two passes of text encoder, we do a single pass. 
     - Text encoder time reduced from 3.54s to 3.06s
     - Generation time per video dropped from 13.84s to 13.38s
    
- JITting Diffusion Loop: The current implementation uses a Python for loop to iterate over diffusion timesteps. This created a "Python Dispatch Wall," resulting in some idle time between consecutive forward passes while the TPU waited for the host CPU to dispatch the next step. We refactored the entire denoising loop to use nnx.scan. 
   - The total diffusion time across 40 steps dropped from 7.84s to 7.28s
   - Generation time per video dropped from 13.38s to 12.5s

### LTX2 Upsampler fix:
- The current LTX2 Upsampler pipeline raises ValueError : blur_down is the name of the submodule in the PyTorch state dict from the Hugging Face checkpoint. In the original PyTorch model, that layer was named blur_down, but in the MaxDiffusion Flax implementation, it was named blur. Because our weight loader didn't rename it, nnx.update tried to update a non-existent blur_down attribute.

### Results

v7x-8

| Version | Execution Time | Status |
| :--- | :--- | :--- |
| **Current Main** | 20.01s | [Video Link](https://storage.mtls.cloud.google.com/prishajain-wan/videos/main_ltx2.mp4) |
| **After Fix** | 12.50s | [Video Link](https://storage.mtls.cloud.google.com/prishajain-wan/videos/updated_ltx2.mp4) |

We also tested WAN I2V pipelines to ensure no regressions are caused there. No quality regression or increased latency was observed.

### DIFF
--- a/src/maxdiffusion/configs/ltx2_video.yml
+++ b/src/maxdiffusion/configs/ltx2_video.yml
@@ -2,11 +2,17 @@
 hardware: 'tpu'
 skip_jax_distributed_system: False
 attention: 'flash'
-a2v_attention_kernel: 'flash'
+a2v_attention_kernel: 'dot_product'
 v2a_attention_kernel: 'dot_product'
 attention_sharding_uniform: True 
 precision: 'bf16'
+
+# For scanning transformer layers
 scan_layers: True
+
+# For scanning diffusion loop
+scan_diffusion_loop: True
+
 names_which_can_be_saved: []
 names_which_can_be_offloaded: []
 remat_policy: "NONE"
--- a/src/maxdiffusion/models/attention_flax.py
+++ b/src/maxdiffusion/models/attention_flax.py
@@ -287,11 +287,11 @@ def _tpu_flash_attention(
 ) -> jax.Array:
   """TPU Flash Attention"""
 
-  block_sizes = _select_flash_block_sizes(query, key, flash_block_sizes, dtype, attention_kernel)
   num_context_shards = mesh.shape["context"]
   query, orig_q_seq_len = _reshape_data_for_flash(query, heads, num_context_shards)
   key, _ = _reshape_data_for_flash(key, heads, num_context_shards)
   value, _ = _reshape_data_for_flash(value, heads, num_context_shards)
+  block_sizes = _select_flash_block_sizes(query, key, flash_block_sizes, dtype, attention_kernel)
 
   q_axis_names = nn.logical_to_mesh_axes(axis_names_q)
   kv_axis_names = nn.logical_to_mesh_axes(axis_names_kv)
@@ -892,7 +892,7 @@ class NNXSimpleFeedForward(nnx.Module):
         dtype=dtype,
         param_dtype=weights_dtype,
         precision=precision,
-        kernel_init=nnx.with_partitioning(nnx.initializers.lecun_normal(), ("embed", "mlp")),
+        kernel_init=nnx.with_partitioning(nnx.initializers.lecun_normal(), (None, "mlp")),
         bias_init=nnx.with_partitioning(nnx.initializers.zeros, ("mlp",)),
     )
     self.act = get_activation(activation_fn)
@@ -904,8 +904,8 @@ class NNXSimpleFeedForward(nnx.Module):
         dtype=dtype,
         param_dtype=weights_dtype,
         precision=precision,
-        kernel_init=nnx.with_partitioning(nnx.initializers.lecun_normal(), ("mlp", "embed")),
-        bias_init=nnx.with_partitioning(nnx.initializers.zeros, ("embed",)),
+        kernel_init=nnx.with_partitioning(nnx.initializers.lecun_normal(), ("mlp", None)),
+        bias_init=nnx.with_partitioning(nnx.initializers.zeros, (None,)),
     )
 
   def __call__(self, hidden_states: Array) -> Array:
--- a/src/maxdiffusion/models/ltx2/attention_ltx2.py
+++ b/src/maxdiffusion/models/ltx2/attention_ltx2.py
@@ -20,6 +20,7 @@ import jax
 import jax.numpy as jnp
 from ... import common_types
 from ..attention_flax import NNXAttentionOp
+from maxdiffusion.tpu_utils import get_tpu_type, TpuType
 
 Array = common_types.Array
 Mesh = common_types.Mesh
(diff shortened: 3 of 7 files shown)


src/maxdiffusion/configs/ltx2_video.yml
```diff
@@ -2,7 +2,7 @@
 hardware: 'tpu'
 skip_jax_distributed_system: False
 attention: 'flash'
-a2v_attention_kernel: 'flash'
+a2v_attention_kernel: 'dot_product'
 v2a_attention_kernel: 'dot_product'
```
Perseus14, line 6: In this current setup, is the following understanding correct?

- audio to video cross attention is `dot_product`
- video to audio cross attention is `dot_product`
- self-attention is `flash`

What about the below? Do we need to have a separate config for them?

- text to video cross attention
- text to audio cross attention

src/maxdiffusion/models/ltx2/attention_ltx2.py
Perseus14: Since `qkv_kernel` is changed to `(None, "heads")`, should we do the same for the `out_kernel`? Maybe change it to `("heads", None)` and `out_bias` to `(None,)`

src/maxdiffusion/pipelines/ltx2/ltx2_pipeline.py
```diff
@@ -1543,3 +1539,120 @@ def transformer_forward_pass(
...
+
+  # Initial carry
+  initial_carry = (latents_jax, audio_latents_jax, scheduler_state)
+
+  # Run scan
+  final_carry, _ = nnx.scan(
+      scan_body,
+      in_axes=(nnx.Carry, 0, None),
+      out_axes=(nnx.Carry, 0),
+  )(initial_carry, timesteps_jax, transformer)
```
Perseus14, line 1726: We have a config param `scan_layers` (which regulates transformer blocks, not the iterative pipeline loop) as well as this `nnx.scan` diffusion loop. Could this change confuse what the `scan_layers` config controls for developers? Perhaps we can add a comment to the `scan_layers` config that this controls only transformer blocks 

@entrpn

src/maxdiffusion/configs/ltx2_video.yml
prishajain1: These happen in dot product. Since the `can_use_falsh_attention` boolean variable only uses flash attention when alk k,q,v have seq length > flash_min_seq_length (which in our case is 4096), we are defaulting to using standard dot product attn. This is expected and so I think we don't need a separate parameter for them and can allow them to be controlled by flash_min_seq_length parameter as present in other models as well.

src/maxdiffusion/models/ltx2/attention_ltx2.py
prishajain1: Updated and tested, the all-reduce decreased slightly and latency dropped from 12.5s to 12.45s

src/maxdiffusion/pipelines/ltx2/ltx2_pipeline.py
prishajain1: I have added a new parameter called 'scan_diffusion_loop` to the config file, which gives us the option to use nnx.scan over the diffusion loop as well.  Also added comments in config file explaining which which param does what.

src/maxdiffusion/models/ltx2/attention_ltx2.py
```diff
@@ -349,23 +349,37 @@ def __init__(
...
+      out_bias_sharding_spec: Optional[tuple] = None,
   ):
     self.heads = heads
     self.rope_type = rope_type
     self.dim_head = dim_head
     self.inner_dim = dim_head * heads
     self.dropout_rate = dropout
 
+    # Auto-detect hardware for sharding specs if not overridden
+    device_kind = jax.devices()[0].device_kind
```
Perseus14: Can we move this to `max_utils`?

prishajain1: Moved it to `tpu_utils.py`

src/maxdiffusion/models/ltx2/attention_ltx2.py
```diff
@@ -349,23 +350,37 @@ def __init__(
...
     self.rope_type = rope_type
     self.dim_head = dim_head
     self.inner_dim = dim_head * heads
     self.dropout_rate = dropout
 
+    # Auto-detect hardware for sharding specs if not overridden
+    tpu_type = get_tpu_type()
+    is_ironwood = tpu_type == TpuType.TPU_7X
+
+    if qkv_sharding_spec is None:
```
entrpn, line 370: Can you add a comment why this if statement is here. If it only has been tested with trillium and Ironwood, can you state that in the comment for clarification.

entrpn (approved):

src/maxdiffusion/models/attention_flax.py
```diff
@@ -892,7 +892,7 @@ def __init__(
         dtype=dtype,
         param_dtype=weights_dtype,
         precision=precision,
-        kernel_init=nnx.with_partitioning(nnx.initializers.lecun_normal(), ("embed", "mlp")),
+        kernel_init=nnx.with_partitioning(nnx.initializers.lecun_normal(), (None, "mlp")),
```
csgoogle, line 895: unrelated change?

Same below

PR #388, opened 2026-04-22
Fix VAE decoding error and use base 2 and experimental scheduler

eltsai:
This PR contains following changes:

1. Fixed a VAE decoding issue
2. Added config parameters: `use_base2_exp` and `use_experimental_scheduler`. When set as True, it give us about 8% speedup. See example videos and analysis in b/505444312
3. Removed unnecessary debug message `print("is_norm_path: ", path)` and warning `Warning, batch dimension should be shardable among the devices in data and context axis, batch dimension: 2, devices_in_data_context: 8` -> we calculated devices_in_data_context wrong
4. (Had this in previous branch, I think I removed it by mistake during some merges): pass `profiler_steps` to the inference code, so that instead of doing xprof on the entire generation, we only profile the steps defined.
5. Updated README.md with new best e2e time - now tokamax_flash is better

### DIFF
--- a/README.md
+++ b/README.md
@@ -628,16 +628,16 @@ To generate images, run the following command:
 We added ring attention support for Wan models. Below are the stats for one `720p` (81 frames) video generation (with CFG DP):
 | Accelerator |  Model | Attention Type | Inference Steps | Sharding | e2e Generation Time |
 | -- | -- | -- | -- | -- | -- | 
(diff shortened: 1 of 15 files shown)


entrpn (dismissed):

Perseus14:
LGTM!

Tested on 7x-8 and the video is now clear!

~189s with `tokamax_flash` and XLA flags

Perseus14 (approved):

entrpn (approved):

PR #396, opened 2026-05-03
fix: enable ulysses sharding for custom kernels and improve scaling precision

Perseus14:
# Description

This PR introduces two small but important fixes to the Ulysses attention implementation:
1. **Enable Ulysses sharding for custom kernels**: In `src/maxdiffusion/pyconfig.py`, the check for Ulysses attention is changed from `attention == "ulysses"` to `"ulysses" in attention`. This ensures that custom attention implementations that include "ulysses" in their identifier (e.g., `custom_ulysses`) will correctly trigger Ulysses sequence sharding instead of falling back to default sharding strategies.
3. **Ensure correct padding**: In `src/maxdiffusion/models/attention_flax.py`, padding was done for `query` variable but `query_scaled` was used in the attention calculation. This fixes it and ensures the padded variable is used in the attention calculation. 
2. **Improve scaling precision**: In `src/maxdiffusion/models/attention_flax.py`, the hardcoded constant `1.44269504` used to scale queries for base-2 exponentiation is replaced with `math.log2(math.e)`. This provides better precision and makes the intent of the code clearer.

## Generation Time
- **Main**: ~420s
- **Branch**: ~140s

### DIFF
--- a/src/maxdiffusion/models/attention_flax.py
+++ b/src/maxdiffusion/models/attention_flax.py
@@ -35,6 +35,7 @@ from ..kernels import custom_splash_attention as custom_splash
 from . import quantizations
 from .modeling_flax_utils import get_activation
 
+LOG2E = math.log2(math.e)
 
 Array = common_types.Array
 Mesh = common_types.Mesh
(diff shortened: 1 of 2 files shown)


eltsai:
Thanks for fixing this @Perseus14 ! I tested on v6e (don't have v7x for now due to capacity) and the generation time is 204 sec:
```
==================================================
  TIMING SUMMARY
==================================================
  Load (checkpoint):      97.3s
  Compile:               229.8s
  ────────────────────────────────────────
  Inference:             204.0s
  Conditioning:           11.8s
  Denoise Total:         183.6s
  VAE Decode:              8.6s
==================================================
```
I think the generation time is expected, because on v7x we are seeing 28% speed boost (140 vs 194.4). From go/wan-dashboard the e2e gen time is 322 sec, the speed gain is about 36% percent.

eltsai (approved):

PR #405, opened 2026-05-13
Fixes import errors for huggingface_hub.utils

martinarroyo:
In some platforms the import is slightly different due to multiple versions of the library being present. This adds a try-catch clause to handle those scenarios.


Perseus14 (approved):

martinarroyo:
This might not be necessary thanks to some changes on the platform, halting merging for now

martinarroyo:
After some testing with the latest updates this is no longer necessary.

PR #406, opened 2026-05-15
feat: add pretrained orbax cache for fast WAN inference loads

csgoogle:
## Pretrained Orbax cache for fast WAN inference loads

Caches WAN weights as an Orbax checkpoint so inference can skip the slow diffusers load on repeat runs (~10× faster). The first run loads from diffusers and writes the cache; subsequent runs restore directly from it. Works with both local paths and GCS buckets.

### Changes
- Added a new `pretrained_orbax_dir` config option to all 6 WAN configs (empty by default, so the feature is off unless set). Accepts a local path or a `gs://` bucket path.
- Load priority: training checkpoint → pretrained cache → diffusers (the cache is populated automatically on a miss).
- `generate_wan` and `generate_wan_animate` now load with `use_pretrained_cache=True`.
- Refactored the shared Orbax logic into the `WanCheckpointer` base class; subclasses now only declare `model_name`, `checkpoint_state_item_names`, and `pretrained_state_sources`.
- Added an animate checkpointer along with tests covering the cache hit / miss / save paths.

image

### DIFF
--- a/src/maxdiffusion/checkpointing/wan_checkpointer.py
+++ b/src/maxdiffusion/checkpointing/wan_checkpointer.py
@@ -15,8 +15,15 @@ limitations under the License.
 """
 
 from abc import ABC, abstractmethod
+import json
 from typing import Optional, Tuple
-from maxdiffusion.checkpointing.checkpointing_utils import (create_orbax_checkpoint_manager)
+import jax
+from flax import nnx
+from maxdiffusion.checkpointing.checkpointing_utils import (
+    add_sharding_to_struct,
+    create_orbax_checkpoint_manager,
+    get_cpu_mesh_and_sharding,
+)
 from ..pipelines.wan.wan_pipeline_2_1 import WanPipeline2_1
 from ..pipelines.wan.wan_pipeline_2_2 import WanPipeline2_2
 from ..pipelines.wan.wan_pipeline_i2v_2p1 import WanPipelineI2V_2_1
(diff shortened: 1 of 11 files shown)


src/maxdiffusion/checkpointing/wan_checkpointer_2_2.py
```diff
@@ -83,20 +89,100 @@ def load_diffusers_checkpoint(self):
     pipeline = WanPipeline2_2.from_pretrained(self.config)
     return pipeline
 
+  def _get_pretrained_orbax_dir(self) -> str:
+    return getattr(self.config, "pretrained_orbax_dir", "")
```
Perseus14: Do we need this function? We could just directly use `getattr(self.config, "pretrained_orbax_dir", "")` as it is only needed in one place. Also should the default value be `None` instead of `""`

src/maxdiffusion/checkpointing/wan_checkpointer_i2v_2p2.py
```diff
@@ -83,20 +89,100 @@ def load_diffusers_checkpoint(self):
     pipeline = WanPipelineI2V_2_2.from_pretrained(self.config)
     return pipeline
 
+  def _get_pretrained_orbax_dir(self) -> str:
+    return getattr(self.config, "pretrained_orbax_dir", "")
```
Perseus14: Same as above

Perseus14 (commented):
Can you also add this for the other WAN models?

csgoogle:
> Can you also add this for the other WAN models?

done

Perseus14:
Would this change have an impact on any training runs?

csgoogle:
> Would this change have an impact on any training runs?

It won't there are unit tests covering the training load, also the flag for the change only get's enabled if it's `true`

csgoogle:
> Would this change have an impact on any training runs?

Made it simple, now we are not touching training code.

Perseus14 (approved):

Perseus14:
LGTM! Thanks for proactively reducing one of the development pain points!

PR #407, opened 2026-05-15
Fix VAE spatial sharding dynamic calculation bug in Wan pipeline.

ninatu:
Previously, setting `vae_spatial: -1` in the config (intended to trigger dynamic calculation of the VAE spatial sharding axis size) was ineffective because `pyconfig.py` prematurely overrode any `-1` or missing `vae_spatial` value to `1`.

Furthermore, the dynamic calculation formula in `wan_pipeline.py` (`vae_spatial = (2 * total_devices) // dp_size`) was not robust. On single-device runs (where `total_devices=1` and `dp_size=1`) or  configurations with odd data parallel (DP) sizes, it would calculate a `vae_spatial` value (e.g., 2) that does not divide `total_devices`, failing the mesh validation assertion.

### DIFF
--- a/src/maxdiffusion/pipelines/wan/wan_pipeline.py
+++ b/src/maxdiffusion/pipelines/wan/wan_pipeline.py
@@ -672,11 +672,8 @@ class WanPipeline:
     vae_spatial = getattr(config, "vae_spatial", -1)
     total_devices = math.prod(devices_array.shape)
 
-    if vae_spatial <= 0:
-      dp_size = mesh.shape.get("data", 1)
-      if dp_size == -1 or dp_size == 0:
-        dp_size = 1
-      vae_spatial = (2 * total_devices) // dp_size
+    if vae_spatial == -1:
+      vae_spatial = total_devices
 
     assert (
         total_devices % vae_spatial == 0
--- a/src/maxdiffusion/pyconfig.py
+++ b/src/maxdiffusion/pyconfig.py
@@ -281,8 +281,8 @@ class _HyperParameters:
         raw_keys["global_batch_size_to_train_on"],
     ) = _HyperParameters.calculate_global_batch_sizes(raw_keys["per_device_batch_size"])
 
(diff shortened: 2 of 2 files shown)


eltsai (dismissed):

src/maxdiffusion/pipelines/wan/wan_pipeline.py
```diff
@@ -676,7 +676,15 @@ def _create_common_components(cls, config, vae_only=False, i2v=False):
...
-      vae_spatial = (2 * total_devices) // dp_size
+      non_dp_size = total_devices // dp_size
+      # VAE activations are huge; we want to maximize VAE spatial sharding (vae_spatial).
+      # Ideally, we want 2x the Transformer's model parallel size (2 * non_dp_size).
+      # However, vae_spatial must divide total_devices, which requires dp_size to be even.
+      # If dp_size is odd (e.g., single-device runs), we fallback to 1x (non_dp_size).
+      if dp_size % 2 == 0:
+        vae_spatial = 2 * non_dp_size
+      else:
+        vae_spatial = non_dp_size
```
Perseus14: Do we need `dp` info for `vae_spatial`? Can we do this instead?

```
    vae_spatial = getattr(config, "vae_spatial", -1)
    total_devices = math.prod(devices_array.shape)

    if vae_spatial == -1:
      vae_spatial = total_devices
```

ninatu: I don't think it's critical, I just kept as it was before. @eltsai, could you comment on this? I think this heuristic was introduced in your commit

eltsai: no we don't need dp info. I was trying to set it assuming we were using dp cfg, but I agree with @Perseus14's snippet, that is more neat

ninatu: thanks, fixed!

Perseus14 (approved):

PR #409, opened 2026-05-15
Fix XProf GCS upload bug in max_utils.Profiler

eltsai:
Creating this PR based on [CL](https://critique.corp.google.com/cl/916091536):

This change fixes a bug where XProf traces were not uploaded to GCS when the profiler was used manually via `start()` and `stop()` calls, which is the case in most MaxDiffusion trainers.

The GCS upload logic was previously located in `Profiler.__exit__` (used by the context manager), but not in `Profiler.stop()`. This change moves the upload logic to `stop()` and guards it with `_jax_profiler_enabled(self.config)` to ensure it only runs when JAX profiling was active.

Also:
- Added unit tests in `profiler_test.py` to verify GCS upload for both manual and context manager usage.
- Added `profiler_test` target to `BUILD`.

### DIFF
--- a/src/maxdiffusion/generate_wan.py
+++ b/src/maxdiffusion/generate_wan.py
@@ -251,6 +251,10 @@ def run(config, pipeline=None, filename_prefix="", commit_hash=None):
 
   s0 = time.perf_counter()
 
+  # Disable profiler for the first two runs to avoid duplicate uploads
+  original_enable_profiler = config.enable_profiler if "enable_profiler" in config.get_keys() else False
+  config.get_keys()["enable_profiler"] = False
+
   # Using global_batch_size_to_train_on so not to create more config variables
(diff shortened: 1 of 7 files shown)


eltsai:
Also disabled profiler for the first two runs. The xprof functions is as expected now:
```
gcloud storage ls gs://elisatsai-wan-maxdiffusion/wan22/8tpu/ulysses-custom-test-xprof3/ulysses-custom-test-xprof3/tensorboard/plugins/profile/2026_05_15_22_27_45/t1v-n-cd58a6ba-w-0.xplane.pb

gs://elisatsai-wan-maxdiffusion/wan22/8tpu/ulysses-custom-test-xprof3/ulysses-custom-test-xprof3/tensorboard/plugins/profile/2026_05_15_22_27_45/t1v-n-cd58a6ba-w-0.xplane.pb
```

Perseus14 (approved):

PR #411, opened 2026-05-18
Optimize Wan VAE: Add Temporal Chunking, On-TPU Postprocessing, Static JIT Support, and Configurable Dtypes

Perseus14:
## Overview
This pull request introduces significant memory, compilation, and bandwidth optimizations for the Wan (T2V, I2V, Vace and Animate) VAE encoding and decoding execution graphs on TPUs. By enforcing static shapes and moving postprocessing operations on-device, this PR significantly improves scaling efficiency and execution robustness.

### 1. VAE Temporal Chunking & Static Compilation
* Introduced configurable chunk size parameters (`vae_decode_chunk` and `vae_encode_chunk`) in the base configuration files.
* Added static temporal zero-padding prior to `jax.lax.scan` loops and precise output trimming post-scan. This guarantees uniform static shapes across iterations, allowing seamless static JIT/XLA compilation without dynamic shape fallbacks or recompilation overhead.

### 2. On-TPU Output Quantization
* Transitioned the final video post-processing stages (range rescaling to `[0, 1]`, clipping, and scaling to `uint8`) directly into the JAX/TPU execution graph across all Wan pipelines (`WanPipeline`, `WanPipelineI2V`, and `WanAnimatePipeline`).
* Gathering sharded arrays as `uint8` instead of floating-point cuts device-to-host memory interconnect transfer bandwidth by up to 4x.
* Completely eliminated the need for `torch.Tensor` conversions and PyTorch processor overhead during generation output assembly.

### 3. Distributed Sharding Robustness
* Standardized VAE 1D spatial sharding rules to use `P("redundant", None, None, "vae_spatial", None)`. Explicitly marking the batch axis as redundant prevents unintended cross-mesh synchronization across replicas under XLA SPMD partitioning.
* Added defensive dimension existence checks when accessing mesh properties in `attention_flax.py` to support diverse hardware topologies gracefully.

### 4. Configurable Activation & Weight Precision
* Integrated robust type casting in `pyconfig.py` for user-defined numerical precision (`vae_dtype`, `vae_weights_dtype`, and `scheduler_dtype`), defaulting VAE runtime memory down to 16-bit precision to save 50% High Bandwidth Memory (HBM) capacity.

| Model | Variant | BaselIne Generation Time | Current Generation Time |
|---|---|---|---|
| WAN2.2 | T2V | 132.2s | 130.6s |
| WAN2.2 | I2V | 133.3s | 132.3s |
| WAN2.1 | T2V | 132.1s | 130.6s |
| WAN2.1 | I2V | 142.7s | 141.6s |

**Conclusion**: No visual change from baseline across all tested variants.

### DIFF
--- a/src/maxdiffusion/configs/base_wan_14b.yml
+++ b/src/maxdiffusion/configs/base_wan_14b.yml
@@ -47,9 +47,23 @@ text_encoder_dtype: 'float32'
 # Whether to compile the text_encoder with torch.compile
 compile_text_encoder: False
 
+# Maximum sequence length for the text encoder
+max_sequence_length: 512
+
+vae_weights_dtype: 'float32'
+vae_dtype: 'float32'
+scheduler_dtype: 'float32'
+
 # Replicates vae across devices instead of using the model's sharding annotations for sharding.
 replicate_vae: False
-vae_spatial: -1 # default to total_device * 2 // (dp)
+
+# Chunk size for VAE decode scan. Increase to improve decode time at the cost of memory.
+vae_decode_chunk: 1
+
+# Chunk size for VAE encode scan. (num_input_frames - 1) must be divisible by this value.
+# Increase to improve encode time at the cost of memory.
+vae_encode_chunk: 4
+vae_spatial: -1
 
 # matmul and conv precision from https://jax.readthedocs.io/en/latest/jax.lax.html#jax.lax.Precision
 # Options are "DEFAULT", "HIGH", "HIGHEST"
--- a/src/maxdiffusion/models/wan/autoencoder_kl_wan.py
+++ b/src/maxdiffusion/models/wan/autoencoder_kl_wan.py
@@ -119,9 +119,7 @@ class WanCausalConv3d(nnx.Module):
     )
 
   def __call__(self, x: jax.Array, cache_x: Optional[jax.Array] = None, idx=-1) -> jax.Array:
-    # Sharding Width (index 3)
-    # Spec: (Batch, Time, Height, Width, Channels)
-    spatial_sharding = NamedSharding(self.mesh, P(None, None, None, "vae_spatial", None))
+    spatial_sharding = NamedSharding(self.mesh, P("redundant", None, None, "vae_spatial", None))
     x = jax.lax.with_sharding_constraint(x, spatial_sharding)
 
     current_padding = list(self._causal_padding)
@@ -159,26 +157,30 @@ class WanRMS_norm(nnx.Module):
       images: bool = True,
       eps: float = 1e-6,
       use_bias: bool = False,
+      dtype: jnp.dtype = jnp.float32,
+      weights_dtype: jnp.dtype = jnp.float32,
   ):
     broadcastable_dims = (1, 1, 1) if not images else (1, 1)
     shape = (dim, *broadcastable_dims) if channel_first else (dim,)
     self.eps = eps
     self.channel_first = channel_first
     self.scale = dim**0.5
+    self.dtype = dtype
     # Initialize gamma as parameter
-    self.gamma = nnx.Param(jnp.ones(shape))
+    self.gamma = nnx.Param(jnp.ones(shape, dtype=weights_dtype))
     if use_bias:
-      self.bias = nnx.Param(jnp.zeros(shape))
+      self.bias = nnx.Param(jnp.zeros(shape, dtype=weights_dtype))
     else:
       self.bias = 0
 
   def __call__(self, x: jax.Array) -> jax.Array:
+    x = x.astype(self.dtype)
     normalized = jnp.linalg.norm(x, ord=2, axis=(1 if self.channel_first else -1), keepdims=True)
     normalized = x / jnp.maximum(normalized, self.eps)
     normalized = normalized * self.scale * self.gamma
     if self.bias:
-      return normalized + self.bias.value
-    return normalized
+      return (normalized + self.bias.value).astype(self.dtype)
+    return normalized.astype(self.dtype)
 
 
 class WanUpsample(nnx.Module):
@@ -254,6 +256,7 @@ class WanResample(nnx.Module):
   ):
     self.dim = dim
     self.mode = mode
+    self.dtype = dtype
     self.time_conv = nnx.data(None)
 
     if mode == "upsample2d":
@@ -340,6 +343,7 @@ class WanResample(nnx.Module):
     # Input x: (N, D, H, W, C), assume C = self.dim
     b, t, h, w, c = x.shape
     assert c == self.dim
+    x = x.astype(self.dtype)
 
     if self.mode == "upsample3d":
       if feat_cache is not None:
@@ -353,7 +357,7 @@ class WanResample(nnx.Module):
             # cache last frame of last two chunk
             cache_x = jnp.concatenate([jnp.expand_dims(feat_cache[idx][:, -1, :, :, :], axis=1), cache_x], axis=1)
           if cache_x.shape[1] < 2 and feat_cache[idx] is not None and isinstance(feat_cache[idx], RepSentinel):
-            cache_x = jnp.concatenate([jnp.zeros(cache_x.shape), cache_x], axis=1)
+            cache_x = jnp.concatenate([jnp.zeros(cache_x.shape, dtype=cache_x.dtype), cache_x], axis=1)
           if isinstance(feat_cache[idx], RepSentinel):
             x = self.time_conv(x)
           else:
@@ -402,7 +406,9 @@ class WanResidualBlock(nnx.Module):
     self.nonlinearity = get_activation(non_linearity)
 
     # layers
-    self.norm1 = WanRMS_norm(dim=in_dim, rngs=rngs, images=False, channel_first=False)
+    self.norm1 = WanRMS_norm(
+        dim=in_dim, rngs=rngs, images=False, channel_first=False, dtype=dtype, weights_dtype=weights_dtype
+    )
     self.conv1 = WanCausalConv3d(
         rngs=rngs,
         in_channels=in_dim,
@@ -414,7 +420,9 @@ class WanResidualBlock(nnx.Module):
         weights_dtype=weights_dtype,
         precision=precision,
     )
-    self.norm2 = WanRMS_norm(dim=out_dim, rngs=rngs, images=False, channel_first=False)
+    self.norm2 = WanRMS_norm(
+        dim=out_dim, rngs=rngs, images=False, channel_first=False, dtype=dtype, weights_dtype=weights_dtype
+    )
     self.conv2 = WanCausalConv3d(
         rngs=rngs,
         in_channels=out_dim,
@@ -488,7 +496,8 @@ class WanAttentionBlock(nnx.Module):
       precision: jax.lax.Precision = None,
   ):
     self.dim = dim
-    self.norm = WanRMS_norm(rngs=rngs, dim=dim, channel_first=False)
+
+    self.norm = WanRMS_norm(rngs=rngs, dim=dim, channel_first=False, dtype=dtype, weights_dtype=weights_dtype)
     self.to_qkv = nnx.Conv(
         in_features=dim,
         out_features=dim * 3,
@@ -510,6 +519,7 @@ class WanAttentionBlock(nnx.Module):
         precision=precision,
     )
 
+  @jax.named_scope("WanVAEAttentionBlock")
   def __call__(self, x: jax.Array, feat_cache=None, feat_idx=0):
     identity = x
     batch_size, time, height, width, channels = x.shape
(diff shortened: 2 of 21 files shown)


src/maxdiffusion/pipelines/wan/wan_pipeline_animate.py
```diff
@@ -1041,5 +1041,13 @@ def __call__(
       return seg_latents
 
     # Postprocess to [0, 1] numpy.
-    video_torch = torch.from_numpy(np.array(video_cf.astype(jnp.float32))).to(torch.bfloat16)
-    return self.video_processor.postprocess_video(video_torch, output_type="np")
+    with self.vae_mesh, nn_partitioning.axis_rules(self.vae_logical_axis_rules):
```
ninatu, line 1049: Would it be possible to use self._decode_latents_to_video here?

src/maxdiffusion/pipelines/wan/wan_pipeline_2_1.py
```diff
@@ -176,7 +176,7 @@ def __call__(
     trace["denoise_total"] = time.perf_counter() - t_denoise_start
 
     t_decode_start = time.perf_counter()
-    video = self._decode_latents_to_video(latents)
+    video = self._decode_latents_to_video(latents, trace=trace)
```
ninatu, line 181: Could it be also added in Wan-VACE pipeline, that all pipelines are aligned? Now all pipelines except VACE return np.array in [0, 225] and only VACE returns in [0. 1]

src/maxdiffusion/configs/base_wan_14b.yml
```diff
@@ -41,10 +41,19 @@ revision: ''
 weights_dtype: 'bfloat16'
 # This sets the layer's dtype in the model. Ex: nn.Dense(dtype=activations_dtype)
 activations_dtype: 'bfloat16'
+vae_weights_dtype: 'bfloat16'
```
ninatu: Could default vae_weights_dtype and vae_dtype be kept to float32 as in the original models? Since bfloat16  might add unexpected artifacts. It refers to all configs

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -1111,64 +1141,98 @@ def _encode(self, x: jax.Array, feat_cache: AutoencoderKLWanCache):
...
     t = x.shape[1]
-    iter_ = 1 + (t - 1) // 4
+    CHUNK_SIZE = self.vae_encode_chunk
+    if CHUNK_SIZE == -1:
+      CHUNK_SIZE = max(1, t - 1)
+      assert (
+          CHUNK_SIZE % self.temporal_downsample_factor == 0
+      ), f"When vae_encode_chunk is -1, the input sequence length - 1 ({CHUNK_SIZE}) must be a multiple of the temporal downsampling factor ({self.temporal_downsample_factor})."
+    # Number of chunk iterations: 1 for init frame, then ceil((t-1)/CHUNK_SIZE) for the rest
+    iter_ = 1 + ((t - 1 + CHUNK_SIZE - 1) // CHUNK_SIZE) if t > 1 else 1
```
ninatu, line 1151: I'm not sure that this logic is correct, shouldn't it be iter_ = 1 + ((t - 1) // CHUNK_SIZE) ?

Perseus14: Thanks, done!

src/maxdiffusion/configs/base_wan_14b.yml
Perseus14: Done!

src/maxdiffusion/pipelines/wan/wan_pipeline_2_1.py
Perseus14: Done!

src/maxdiffusion/pipelines/wan/wan_pipeline_animate.py
Perseus14: For `WANAnimate` we use `self._decode_segment_to_pixels` as inference runs over multi-frame sequential overlap segments and each segment needs to be decoded immediately into raw pixel space to provide overlap conditioning to the next segment.

To get the same speed and savings, I refactored the final assembly block to use native JAX on-TPU postprocessing.

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -1111,64 +1141,98 @@ def _encode(self, x: jax.Array, feat_cache: AutoencoderKLWanCache):
...
+      local_encoder = nnx.merge(graphdef, state)
+      out_chunk, next_feat_map, _ = local_encoder(chunk, feat_cache=current_feat_map, feat_idx=0)
       out_chunk = jax.lax.with_sharding_constraint(out_chunk, spatial_sharding)
+      next_feat_map = jax.tree_util.tree_map(
+          lambda x: jax.lax.with_sharding_constraint(x, spatial_sharding) if isinstance(x, jax.Array) else x, next_feat_map
+      )
       return next_feat_map, out_chunk
 
     enc_feat_map, out_rest = jax.lax.scan(scan_fn, enc_feat_map, x_scannable)
     # out_rest shape: (iter_-2, B, T', H, W, C) -> transpose back
```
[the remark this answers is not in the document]
Perseus14: I am doing this a few lines later

src/maxdiffusion/pipelines/wan/wan_pipeline.py
```diff
@@ -648,21 +659,27 @@ def prepare_latents_i2v_base(
...
+      video = (video * 255.0).astype(jnp.uint8)
+      video.block_until_ready()
+    if trace is not None:
+      trace["vae_decode_tpu"] = time.perf_counter() - t_vae_tpu_start
 
-    video = jnp.transpose(video, (0, 4, 1, 2, 3))
     video = jax.experimental.multihost_utils.process_allgather(video, tiled=True)
-    video = torch.from_numpy(np.array(video.astype(dtype=jnp.float32))).to(dtype=torch.bfloat16)
-    return self.video_processor.postprocess_video(video, output_type="np")
+    video = np.array(video)
```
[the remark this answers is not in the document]
Perseus14: Yes, downstream changes have been made

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -1111,64 +1141,98 @@ def _encode(self, x: jax.Array, feat_cache: AutoencoderKLWanCache):
...
 
-    spatial_sharding = NamedSharding(self.mesh, P(None, None, None, "vae_spatial", None))
+    spatial_sharding = NamedSharding(self.mesh, P("redundant", None, None, "vae_spatial", None))
 
     def finalize(out, enc_feat_map):
       feat_cache._enc_feat_map = enc_feat_map
       enc = self.quant_conv(out)
-      mu, logvar = enc[:, :, :, :, : self.z_dim], enc[:, :, :, :, self.z_dim :]
-      enc = jnp.concatenate([mu, logvar], axis=-1)
       feat_cache.init_cache()
```
ninatu: is feat_cache.init_cache() needed here? feat_cache.init_cache() is always called in the beginning of the function, therefore it seems like here it should not be needed

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -1111,64 +1141,98 @@ def _encode(self, x: jax.Array, feat_cache: AutoencoderKLWanCache):
...
     t = x.shape[1]
-    iter_ = 1 + (t - 1) // 4
+    CHUNK_SIZE = self.vae_encode_chunk
+    if CHUNK_SIZE == -1:
+      CHUNK_SIZE = max(1, t - 1)
+      assert (
+          CHUNK_SIZE % self.temporal_downsample_factor == 0
+      ), f"When vae_encode_chunk is -1, the input sequence length - 1 ({CHUNK_SIZE}) must be a multiple of the temporal downsampling factor ({self.temporal_downsample_factor})."
+    # Number of chunk iterations: 1 for init frame, then (t-1)//CHUNK_SIZE for the rest
+    iter_ = 1 + ((t - 1) // CHUNK_SIZE) if t > 1 else 1
```
ninatu: Sorry, just double checked, it seems like it should be `1 +  int(math.ceil((t - 1) / CHUNK_SIZE))`?  For the edge case, when the last chunk is not full, right?

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
ninatu: It seems like feat_cache.init_cache() is unnecessary here as well

ninatu (dismissed):

src/maxdiffusion/tests/wan_kv_cache_test.py
```diff
@@ -210,7 +210,7 @@ def mock_load_scheduler(config):
     self.assertEqual(video_with_cache[0].shape, (num_frames, height, width, 3))
 
     # Compare outputs
-    np.testing.assert_allclose(video_no_cache, video_with_cache, rtol=1e-1, atol=0.7)
+    np.testing.assert_allclose(video_no_cache, video_with_cache, rtol=1e-1, atol=180)
```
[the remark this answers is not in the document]
Perseus14: Refactored KVCache test to check MAE instead of checking each pixel difference.

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
Perseus14: Ahh that's right. I will revert to the previous logic instead of importing a new math package

`iter_ = 1 + ((t - 1 + CHUNK_SIZE - 1) // CHUNK_SIZE) if t > 1 else 1`

ninatu: could you maybe add the comment about it? Since it was unclear that  `CHUNK_SIZE - 1` addition for rounding up,  I was confused.

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
Perseus14: @eltsai Could you confirm? I tested by removing it and it works, video is generated with no quality difference.

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -159,26 +157,30 @@ def __init__(
...
   ):
     broadcastable_dims = (1, 1, 1) if not images else (1, 1)
     shape = (dim, *broadcastable_dims) if channel_first else (dim,)
     self.eps = eps
     self.channel_first = channel_first
     self.scale = dim**0.5
+    self.dtype = dtype
     # Initialize gamma as parameter
-    self.gamma = nnx.Param(jnp.ones(shape))
+    self.gamma = nnx.Param(jnp.ones(shape, dtype=weights_dtype))
```
eltsai, line 170: this is great!

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -1111,64 +1139,97 @@ def _encode(self, x: jax.Array, feat_cache: AutoencoderKLWanCache):
...
       x = jnp.transpose(x, (0, 2, 3, 4, 1))
       assert x.shape[-1] == 3, f"Expected input shape (N, D, H, W, 3), got {x.shape}"
+    x = x.astype(self.encoder.conv_in.conv.dtype)
 
     t = x.shape[1]
-    iter_ = 1 + (t - 1) // 4
+    CHUNK_SIZE = self.vae_encode_chunk
+    if CHUNK_SIZE == -1:
+      CHUNK_SIZE = max(1, t - 1)
+      assert (
```
eltsai: Instead of throwing out an exception here, maybe we can do some padding here to make sure it is dividable? Will see if others think this is reasonable.

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -510,21 +519,23 @@ def __init__(
...
-    # qkv = qkv.reshape(batch_size * time, 1, channels * 3, -1)
-    qkv = qkv.reshape(batch_size * time, 1, -1, channels * 3)
-    qkv = jnp.transpose(qkv, (0, 1, 3, 2))
-    q, k, v = jnp.split(qkv, 3, axis=-2)
-    q = jnp.transpose(q, (0, 1, 3, 2))
-    k = jnp.transpose(k, (0, 1, 3, 2))
-    v = jnp.transpose(v, (0, 1, 3, 2))
+    qkv = self.to_qkv(x)  # Output: (N*T, H, W, C * 3)
+    qkv = qkv.reshape(batch_size * time, height * width, 3, channels)
+    q = qkv[:, :, 0, :]  # (B*T, H*W, C)
```
eltsai: Curious if we have numbers showing the exact speedup from this block alone?

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
Perseus14: Done!

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
Perseus14: I tested it and found that the original implementation is much better interms of speed and memory. I have reverted it to the original code

eltsai (approved):

PR #416, opened 2026-06-09
Fix: Plumb vmem limit and bkv_compute_in to the custom kernel

eltsai:
Right now the `vmem_limit_bytes` and  `block_kv_compute_in` from the argument was never passed into the kernel.

This PR plumbs `vmem_limit_bytes` from the `flash_block_sizes` config down to the Mosaic
CompilerParams so the custom splash kernel can run under a raised scoped-VMEM
limit.

- `custom_splash_attention.py`: `vmem_limit_bytes` param through both forwards ->
  `pltpu.CompilerParams`, `make_splash_mha`, `tpu_custom_attention` (+ flash_block_sizes
  read), make_custom_splash_sdpa (+ kwargs read).
- `attention_flax.py`: `read vmem_limit_bytes` in `wrap_ulysses_attention`, pass to
  `make_splash_mha`.
- `max_utils.py`: `CustomFlashBlockSizes` frozen carrier so `vmem_limit_bytes` (and
  `block_kv_compute_in` / `heads_per_tile`) survive config -> kernel instead of being
  dropped by the JAX BlockSizes dataclass.

By raising the vmem to the 64 MB limit, our grid search finds a better config:

sweep1
sweep2

```
export BQ=8448
export BKV=1024
export BKV_COMPUTE=1024
export BKV_COMPUTE_IN=1024
export HEADS_PER_TILE=1
```

E2E result:
```
==================================================
  TIMING SUMMARY
==================================================
  Load (checkpoint):     141.2s
  Compile:               183.3s
  Inference:             126.9s
  ────────────────────────────────────────
  Conditioning:            2.2s
    - VAE Encode:          0.0s
  Denoise Total:         122.6s
  VAE Decode:              2.1s
    - TPU Compute:         1.7s
    - Host Formatting:     0.4s
==================================================
```

Comparing to previous config (`BQ=4864`), the denoising steps is 3.7% faster (122.6 sec vs 127.3 sec).

### DIFF
--- a/src/maxdiffusion/kernels/custom_splash_attention.py
+++ b/src/maxdiffusion/kernels/custom_splash_attention.py
@@ -359,6 +359,7 @@ def _splash_attention_forward(
     kv_seq_len: int | None = None,
     use_base2_exp: bool = True,
     use_experimental_scheduler: bool = False,
+    vmem_limit_bytes: int | None = None,
 ):
   num_q_heads, padded_q_seq_len, head_dim_qk = q.shape
   head_dim_v = v.shape[-1]
@@ -429,6 +430,7 @@ def _splash_attention_forward(
           flags={"XLA_TPU_FORCE_LP_LLO_SCHEDULER": use_experimental_scheduler},
           disable_bounds_checks=True,
           skip_device_barrier=True,
+          vmem_limit_bytes=vmem_limit_bytes,
       ),
       out_shape=out_shapes,
   )(q, k, v)
@@ -446,6 +448,7 @@ def _splash_attention_forward_mhpt(
     kv_seq_len: int | None = None,
     use_base2_exp: bool = True,
     use_experimental_scheduler: bool = False,
+    vmem_limit_bytes: int | None = None,
 ):
   num_q_heads, padded_q_seq_len, head_dim_qk = q.shape
   head_dim_v = v.shape[-1]
@@ -518,6 +521,7 @@ def _splash_attention_forward_mhpt(
           flags={"XLA_TPU_FORCE_LP_LLO_SCHEDULER": use_experimental_scheduler},
           disable_bounds_checks=True,
           skip_device_barrier=True,
+          vmem_limit_bytes=vmem_limit_bytes,
       ),
       out_shape=out_shapes,
   )(q, k, v)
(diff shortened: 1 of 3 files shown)


Perseus14:
Pretty great find @eltsai, awesome work! 

Few questions
 - Could you clarify on the usage of `vmem_limit`? Do we add this param in the config file under flash_block_sizes, if so can we update the configs as well (if it doesn't affect other attention mechanisms)? 
 - Does setting the XLA flag for vmem_limit_bytes (eg: `xla_tpu_scoped_vmem_limit_kib`) affect this value?
 - Can we update the PR title and description as `block_kv_compute_in` seems to be already passed into the kernel?

Perseus14 (approved):

eltsai:
@Perseus14 (not sure why I can't directly reply to the comments)
> 
> * Could you clarify on the usage of `vmem_limit`? Do we add this param in the config file under flash_block_sizes, if so can we update the configs as well (if it doesn't affect other attention mechanisms)?

Yes, it's read from the flash_block_sizes dict in the config, for example.:
```
flash_block_sizes: {
  "block_q": 5888,
  "block_kv": 1024,
  "block_kv_compute": 1024,
  "block_kv_compute_in": 1024,
  "heads_per_tile": 1,
  "vmem_limit_bytes": 67108864   
```
> * Does setting the XLA flag for vmem_limit_bytes (eg: `xla_tpu_scoped_vmem_limit_kib`) affect this value?

No, they're separate knobs at different layers and neither feeds the other. `vmem_limit_bytes` here is the Mosaic per-kernel VMEM budget for the splash kernel specifically (via CompilerParams).
`xla_tpu_scoped_vmem_limit_kib` is a global XLA-level scoped-VMEM limit for XLA-compiled HLO ops.

> * Can we update the PR title and description as `block_kv_compute_in` seems to be already passed into the kernel?

before this commit, `block_kv_compute_in` (and `heads_per_tile`) were silently dropped by `get_flash_block_sizes`, because it converted the dict into the JAX `BlockSizes` dataclass, which has no field for them. The new CustomFlashBlockSizes carrier added here is what actually lets `block_kv_compute_in`, `heads_per_tile`, and `vmem_limit_bytes` survive to the kernel.

PR #418, opened 2026-06-13
Add Stable Diffusion 1.5 support with data-parallel inference

csgoogle:
## Summary
Adds Stable Diffusion 1.5 to MaxDiffusion and makes its inference data parallel.

- **`configs/base15.yml`** (new) — SD 1.5 config. Mirrors `base14.yml` (same architecture, different weights), points at the `stable-diffusion-v1-5` checkpoint, sets `from_pt: True` (upstream ships PyTorch weights only), and defaults to the checkpoint's PNDM / epsilon scheduler.
- **`generate.py`**
  - Build the sampler from the checkpoint's scheduler config via `create_scheduler` (config-driven, mirrors the SDXL path) instead of a hardcoded `FlaxDDIMScheduler`, and iterate the full PNDM schedule (`skip_prk_steps` emits one extra timestep).
  - Shard the latent batch over the `data` axis using `with_sharding_constraint` + a batch-sharded `out_shardings`, so GSPMD propagates data-parallelism through the whole UNet/VAE instead of replicating the entire batch on every device. A single `get_batch_sharding` helper is the source of truth and replicates for sub-device batches (`per_device_batch_size < 1`).
  - `override_scheduler_config` is now tolerant of scheduler configs that omit keys (e.g. SD 1.5's older PNDM config).

## Why
The previous generate path declared a `data` mesh axis but the program ran fully replicated (`device_put` inside the jit is a weak hint GSPMD ignored), so each chip recomputed the whole batch. Forcing the batch sharding makes inference genuinely data parallel.

## Performance
TPU7x (8 chips), SD 1.5 1024px, 20-step PNDM, 8 images: **0.130s** . Scales cleanly to larger batches.

## Test plan
- [x] Generates coherent images matching the prompts.

Prompt: "A cinematic photo of a glass greenhouse on a snowy mountain"
image
image
image

python src/maxdiffusion/generate.py \
  src/maxdiffusion/configs/base15.yml \
  hardware=tpu \
  run_name=sd15_branch_1024_20 \
  output_dir=/mnt/disks/data/maxdiffusion_outputs \
  jax_cache_dir=/mnt/disks/data/maxdiffusion/.jax_cache/sd15_branch_1024_20 \
  resolution=1024 \
  num_inference_steps=20


mbohlool (changes requested):
ignore this comment.

mbohlool (approved):
LGTM

mbohlool (approved):

entrpn (approved):

Issue #423, opened 2026-06-21
LTX-2.3 pipeline: native SIGSEGVs on tpu7x


tsvikas:
> [!IMPORTANT]
> **Update (2026-06-22): the 1024x1536 denoise compile is itself unreliable; the SIGSEGV looks host/run-dependent, not LoRA- or perturbation-specific.**
> Controlled re-testing changes the picture below. The "1024x1536 known-good" baseline referenced throughout is **not reliable**: the identical config (no LoRA, compile-cache off) produced a clip on one VM, but **segfaulted at the denoise compile on a second VM on every attempt (5/5)** — a controlled repeat confirmed it's not run-to-run flaky on that host. Implications:
> - **2b is not a distinct trigger** — the same crash happens with **no LoRA** at the same config. The LoRA fuse is clean (0 unmatched keys); generating with it just hits the same base crash.
> - The "1024 works, a perturbation tips it over" framing (2a, 2e, and the Unifying observation) is **suspect**, because the 1024 baseline it's measured against is itself flaky.
> - **v6e (4-chip) ran the same LTX-2.3 model e2e cleanly** (`GEN_EXIT=0`) at 512x768x121 — both no-LoRA and with a colorization LoRA fused (video produced, profiling pass and all). So the failure is specific to the v7 host(s) we hit, not the model/config/pipeline.
> - Working hypothesis: **host/instance-dependent libtpu/XLA instability** in the denoise compile, not tied to a config knob. Same jax/libtpu versions and command across working and failing runs; the difference is the physical VM. (Evidence, not proof — few hosts sampled, and the v6e run differs in mesh/resolution.) Investigation ongoing.

## Summary

Running the LTX-2.3 path (`configs/ltx2_3_video.yml` + the default `dg845/LTX-2.3-Diffusers` weights) on a single tpu7x (Ironwood, 1 chip = 2 cores, 192 GB HBM) hits several native segmentation faults (`Fatal Python error: Segmentation fault`, process exit 139). 
The LTX-2 path (`configs/ltx2_video.yml` + `Lightricks/LTX-2`) on the same box is stable, so these are 2.3-specific.

They fall into two buckets:

- **(1) Bad-config cases that should raise a clear error instead of segfaulting.** A user passing   an unsupported/invalid setting gets an opaque native crash with no actionable message.
- **(2) Opaque compiler crashes with no obvious bad input: these need investigation.** Including the repo's own default config resolution.

Environment: 
[...]

PR #425, opened 2026-06-21
feat(ltx2): LTX-2.3 LoRA support (map gate_logits + prompt_adaln keys, fix example config to 22b)

tsvikas:
## Summary

Adds LoRA support for **LTX-2.3 (22b)** in the LTX-2 pipeline by extending the NNX→Diffusers LoRA key-translation map and clarifying the loader's diagnostics.

- **Map extensions** (`loaders/lora_conversion_utils.py`):
  - `to_gate_logits` for every attention family: self (`attn1`), cross (`attn2`), audio (`audio_attn1`/`audio_attn2`), and cross-modal (`audio_to_video_attn` / `video_to_audio_attn`).
  - `prompt_adaln` / `audio_prompt_adaln` (linear + timestep-embedder) → the model's `prompt_adaln_single` / `audio_prompt_adaln_single`.
- **Loader logging/diagnostics** (`loaders/ltx2_lora_nnx_loader.py`), merge behavior unchanged:
  - a missing LoRA weight name returns early with one clear message (was two unclear logs).
  - a missing `transformer` and a missing `connectors` each get their own message (the latter had none before).
  - a LoRA carrying keys outside the two routed prefixes (`diffusion_model.` / `text_embedding_projection.`) is now warned about. Those keys were previously dropped with no notice.
- **Example config** (`configs/ltx2_3_video.yml`): point the example `lora_config` at the 22b adapter (colorization IC-LoRA) instead of the prior **19b** camera-control entry, which is the wrong model size for the 2.3 base.

## Scope & validation

- All 13 official LTX-2.3-22b IC-LoRAs (plus the Distilled LoRA) load cleanly: every key maps to an emittable model key. Verified by running each LoRA's keys through the translate map; all 13 IC-LoRAs plus the Distilled LoRA were additionally live-fused on the real 22b model (0 unmatched each).
- **pipeline output**: A with-LoRA text2video run completes **end-to-end on v6e** — the colorization IC-LoRA fuses (0 unmatched) and generation produces a video, `GEN_EXIT=0`. The same generation segfaults at the denoise compile on the v7 host we used — but so does the no-LoRA run there, and the bare config ran clean on a different v7 host, so it's a host-dependent base-pipeline instability (#423), not the LoRA. (These IC-LoRAs need a vid2vid/reference path for their intended effect, which this repo lacks, so the v6e clip is a valid with-LoRA generation, not a colorization ablation.)

## Question for maintainers

Today the loader **logs** any keys that route to no merge target. Would you prefer it to **raise** instead (fail loudly on a wrong-prefix / zero-coverage load rather than silently no-op)? Happy to add that if you'd like; left as log-only for now to avoid changing behavior without your call.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/maxdiffusion/configs/ltx2_3_video.yml
+++ b/src/maxdiffusion/configs/ltx2_3_video.yml
@@ -142,18 +142,20 @@ enable_lora: False
 
 # Distilled LoRA
 # lora_config: {
-#   lora_model_name_or_path: ["Lightricks/LTX-2"],
-#   weight_name: ["ltx-2-19b-distilled-lora-384.safetensors"],
+#   lora_model_name_or_path: ["Lightricks/LTX-2.3"],
+#   weight_name: ["ltx-2.3-22b-distilled-lora-384.safetensors"],
 #   adapter_name: ["distilled-lora-384"],
+#   # placeholder - the real value is mixed per-layer ranks: 32/128/256/384
+#   # and the loader reads each layer's REAL rank from the LoRA tensor shapes
 #   rank: [384]
 # }
 
 # Standard LoRA
 lora_config: {
-  lora_model_name_or_path: ["Lightricks/LTX-2-19b-LoRA-Camera-Control-Dolly-In"],
-  weight_name: ["ltx-2-19b-lora-camera-control-dolly-in.safetensors"],
-  adapter_name: ["camera-control-dolly-in"],
-  rank: [32]
+  lora_model_name_or_path: ["Lightricks/LTX-2.3-22b-IC-LoRA-Colorization"],
+  weight_name: ["ltx-2.3-22b-ic-lora-colorization-0.9.safetensors"],
+  adapter_name: ["colorization"],
+  rank: [128]
 }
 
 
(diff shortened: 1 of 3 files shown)


Perseus14:
Looks good!

Tested on 7x-8 with both single LoRA and multi-LoRA weights and output is generated with no LoRA layers being skipped.

Regarding your query, we can leave it as log, no need to raise/fail.

Please squash your commits

tsvikas:
Thanks! 
I need a passing review to "squash and merge". Alternatively, I can squash locally and force push the branch.

Perseus14:
Please squash and force push. I will drop the approval after

tsvikas:
squashed.

Perseus14 (approved):

tsvikas:
seems like Merging is blocked from `Error: Unhandled error: HttpError: Resource not accessible by integration`, which seems like a hiccup. ?

Perseus14:
@tsvikas We have a bot that merges the PR once it's approved internally in Google codebase. I have requested the owners to look into this.

PR #426, opened 2026-06-22
fix(generate_wan): avoid process0-only gate for profiling run

### LINKED ISSUE #435: generate_wan: profiling pass deadlocks on multi-process TPU (profiler gated to process 0 only)
Summary

Running generate_wan.py with enable_profiler=True on a multi-process TPU pod (≥2 JAX processes) hangs indefinitely during the profiling generation pass and eventually aborts with a barrier DEADLINE_EXCEEDED. The warm-up (non-profiled) pass completes normally on all processes; only the profiled pass deadlocks. The bug is invisible on single-process / single-device runs, which is why it wasn't caught earlier.

Filing per the request in #426 to track the root cause with reproduction details.

Environment

- MaxDiffusion: main at commit 9616d1c2c1b2cffab5a42899cf3c333a8e4de3c1 (bug present; introduced in 589c3d5).
- Workload: WAN2.2-I2V, 1920×1088×81, num_inference_steps=8, mesh DP1·FSDP1·CP4·TP4, attention=ulysses_custom.
- Hardware: multi-host TPU pod (4 JAX processes / 16 chips). Originally observed on a v4-16 pod; also reproducible on v6e-16. Any ≥2-process pod triggers it.
- JAX: jax[tpu] (multi-host, one process per host).

Steps to reproduce

On every host of the pod (e.g. gcloud compute tpus tpu-vm ssh <tpu> --worker=all --command=…), run the same invocation so all processes join the mesh:

```
python src/maxdiffusion/generate_wan.py \
  src/maxdiffusion/configs/base_wan_i2v_27b.yml \
  enable_profiler=True \
  num_inference_steps=8 \
  height=1088 width=1920 \
  per_device_batch_size=0.0625 \
  ici_data_parallelism=1 ici_fsdp_parallelism=1 \
  ici_context_parallelism=4 ici_tensor_parallelism=4 \
  attention=ulysses_custom \
  run_name=wan22_profiler_repro \
  output_dir=/tmp/wan_output base_output_directory=/tmp/wan_output
```

Only two ingredients are essential to trigger it: more than one JAX process and enable_profiler=True. Resolution, step count, and mesh are incidental (the values above are just what we ran).

Observed behavior

Warm-up finishes on all processes, then the profiling pass deadlocks and aborts (log from the reproduction on a 4-process / 16-device pod):
```
generation_time: 522.09s        # warm-up ok on all 4 processes
# hung — generation_time_with_profiler never logged

DEADLINE_EXCEEDED: Barrier timed out.
# of tasks that reached the barrier: 3/4.
The first task at the barrier: 0.
Aborted (core dumped)
```
Process 0 blocks inside call_pipeline; processes 1–3 return early, reach the shutdown barrier first (3/4), and the job times out.

Root cause

In generate_wan.run(), the profiling pass is gated on max_utils.profiler_enabled(config). That helper carries an implicit jax.process_index() == 0 guard (via _jax_profiler_enabled / _ml_diagnostics_profiler_enabled), so it returns True only on the coordinator. As a result only process 0 enters the profiling block and calls call_pipeline, which issues collective ops (AllReduce/AllGather) that require every process to participate. Processes 1–N skip the block and return, so the collective never forms a quorum → indefinite hang → barrier timeout.

Expected behavior

With enable_profiler=True, all processes should enter the profiling pass together so the collectives complete, matching the warm-up pass behavior.

Fix

Proposed in #426: gate the profiling pass on the raw original_enable_profiler boolean (saved from config.enable_profiler before warm-up temporarily disables it) instead of profiler_enabled(config). That value is True on all processes, so every process runs the profiling pass. The jax.process_index() == 0 guard inside profiler_enabled is correct for profiler context management and is left untouched. (1-line change; verified end-to-end — profiled pass then completes on all processes and emits the XLA trace.)

edgexyz:
**Root Cause:**
In commit 589c3d5, the profiling pass in `generate_wan.run()` was gated on
`max_utils.profiler_enabled(config)`. That helper includes an implicit
`jax.process_index() == 0` guard (via `_jax_profiler_enabled` /
`_ml_diagnostics_profiler_enabled`) so it returns `True` only on the coordinator.
This causes only process 0 to enter the profiling block and call `call_pipeline`,
while all other processes skip it and return. Since `call_pipeline` issues
collective JAX ops (AllReduce, AllGather, etc.) that require a barrier from every
process, the job hangs indefinitely with no error message. The bug is invisible on
single-device runs, which masked it.

**Fix:**
Fixes #435. Replace the gate with `original_enable_profiler` — the raw boolean saved from
`config.enable_profiler` before the warm-up pass temporarily disables it. This is
`True` on all processes when the user sets `enable_profiler=True`, so every process
enters the profiling pass together and collectives succeed. The
`jax.process_index() == 0` guard inside `profiler_enabled` is correct for profiler
context management and remains untouched.

Verified end-to-end on a 4-process TPU v4 pod (16 devices), WAN2.2-I2V
1920×1088×81, `enable_profiler=True`:

**Before (589c3d5):** warm-up completed, profiling pass deadlocked.
```
generation_time: 522.09s   # warm-up ok on all 4 processes
# hung — generation_time_with_profiler never logged

DEADLINE_EXCEEDED: Barrier timed out.
# of tasks that reached the barrier: 3/4.
The first task at the barrier: 0.
Aborted (core dumped)
```
Process 0 blocked inside `call_pipeline`; processes 1–3 exited and hit the
shutdown barrier first (3/4 reached it → timeout → abort).

**After (f62927f):** all processes complete, XLA trace produced.
```
Warm generation                : count=4, mean=522.26s
Warm generation with profiler  : count=4, mean=530.24s
Trace artifact count           : 2
```

### DIFF
--- a/src/maxdiffusion/generate_wan.py
+++ b/src/maxdiffusion/generate_wan.py
@@ -353,7 +353,7 @@ def run(config, pipeline=None, filename_prefix="", commit_hash=None):
   s0 = time.perf_counter()
   # Restore original profiler setting for the profiling run
   config.get_keys()["enable_profiler"] = original_enable_profiler
-  if max_utils.profiler_enabled(config):
+  if original_enable_profiler:
     # Injecting user requested XLA tracing flags
     xla_flags = os.environ.get("XLA_FLAGS", "")
     new_flags = "--xla_enable_mxu_trace=true --xla_jf_dump_llo_html=true --xla_tpu_enable_llo_profiling=true"


Perseus14:
Hi @edgexyz Could you please create an issue around the gated profiling? Please add details (hardware, environment, command etc...) to replicate this behaviour

edgexyz:
Thanks @Perseus14 — opened #435 with the full replication details: hardware (multi-host TPU pod, 4 processes / 16 chips), environment, the exact generate_wan.py … enable_profiler=True invocation, and the deadlock / DEADLINE_EXCEEDED barrier logs. Happy to add anything else that would help.

Perseus14:
Thanks for additional details and the fix @edgexyz. LGTM!

Perseus14 (approved):

PR #432, opened 2026-06-29
Optimize WAN pipelines: Reduce host overhead, hoist transposes, and fix Cache bugs

Perseus14:
## Overview
This PR introduces several performance optimizations and bug fixes to the Wan 2.1 and 2.2 pipelines (both T2V and I2V), significantly reducing host overhead during the denoising loop and ensuring stability under data-parallel configurations. 

## Key Optimizations
* **Hoisted I2V Tensor Operations:** Pre-transposed and concatenated the `condition` tensors *outside* of the main denoising loop in the I2V pipelines (`wan_pipeline_i2v_2p1.py`, `wan_pipeline_i2v_2p2.py`) to eliminate redundant `jnp.transpose` and `jnp.concatenate` churn on every inference step.
* **Hoisted Timesteps:** Hoisted timestep calculations outside of the loop where possible.

## Bug Fixes & Cleanups
* **CFG Cache + Ulysses Attention Crash Fix:** Added dynamic checks in both Wan 2.1 and Wan 2.2 pipelines to disable `use_cfg_cache` if `batch_size % data_shards != 0`. Previously, `cfg_cache` evaluated only the unconditional branch (dropping the batch size to 1), which caused a `ValueError` in `ulysses_attention`'s `shard_map` if the user configured `data_shards > 1`.
* **Logging Standardization:** Replaced raw `print()` statements with `max_logging.log()` across the pipelines for cleaner distributed logging.
* **Formatting:** Ran `ruff` and `pyink` to clean up linter warnings and unused imports.

### DIFF
--- a/src/maxdiffusion/pipelines/wan/wan_pipeline_2_1.py
+++ b/src/maxdiffusion/pipelines/wan/wan_pipeline_2_1.py
@@ -34,6 +34,7 @@ from ...schedulers.scheduling_unipc_multistep_flax import FlaxUniPCMultistepSche
 import numpy as np
 import time
 from ... import max_utils
+from maxdiffusion import max_logging
 
 
 class WanPipeline2_1(WanPipeline):
@@ -125,8 +126,12 @@ class WanPipeline2_1(WanPipeline):
       magcache_K: Optional[int] = None,
       retention_ratio: Optional[float] = None,
       use_kv_cache: bool = False,
+      output_type: str = "np",
   ):
     config = getattr(self, "config", None)
+    if output_type not in ["np", "latent"]:
+      raise ValueError(f"output_type must be one of ['np', 'latent'], got {output_type}")
+
     if max_sequence_length is None:
       max_sequence_length = getattr(config, "max_sequence_length", 512)
     if magcache_thresh is None:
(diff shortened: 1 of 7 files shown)


entrpn (commented):
are these changes also needed in the wan animate pipelines? @ninatu

src/maxdiffusion/pipelines/wan/wan_pipeline_2_1.py
```diff
@@ -101,6 +102,7 @@ def __call__(
       magcache_K: Optional[int] = None,
       retention_ratio: Optional[float] = None,
       use_kv_cache: bool = False,
+      output_type: str = "pil",
```
ninatu: By default the method returns `np.array`, not `pil`, but correct me if I'm wrong. In general, it would be nice to make checks that `output_type` is within supported output types.

src/maxdiffusion/pipelines/wan/wan_pipeline_2_2.py
```diff
@@ -118,6 +118,7 @@ def __call__(
       use_cfg_cache: bool = False,
       use_sen_cache: bool = False,
       use_kv_cache: bool = False,
+      output_type: str = "pil",
```
ninatu: Same here

ninatu:
@entrpn 
> are these changes also needed in the wan animate pipelines?

not needed, since magcache is not implemented in wan animate and in wan-vace at the moment. But we it would be great to have magcache implemented in wan-vace as well!

src/maxdiffusion/pipelines/wan/wan_pipeline_2_1.py
Perseus14: Done!

Perseus14:
@ninatu @entrpn PTAL

prishajain1 (approved):

PR #433, opened 2026-06-29
Add MagCache inference acceleration for Wan2.2 (T2V + I2V)

HadarIngonyama:
# Add MagCache inference acceleration for Wan2.2 (T2V + I2V)

## Summary

This PR adds [MagCache](https://github.com/Zehong-Ma/MagCache) support to the Wan2.2 dual-transformer pipelines (both T2V and I2V), extending the existing Wan2.1 T2V MagCache support. MagCache skips the transformer blocks and reuses the cached block residual when the accumulated magnitude-ratio error stays below a threshold, using a precalibrated per-step `mag_ratios_base` curve so the skip schedule is deterministic (no data-dependent control flow, TPU/JIT friendly).

Measured speedups vs the dense render: **~1.82× for T2V** and **~1.75× for I2V**, with visually near-indistinguishable output.

## What's included

- **Wan2.2 T2V** (`wan_pipeline_2_2.py`): MagCache skip path for the dual transformer — a single interleaved `mag_ratios_base` curve spanning both the high-noise and low-noise phases, a per-phase forced-compute (retention) zone, and an explicit cached-residual reset at the high→low transformer boundary.
- **Wan2.2 I2V** (`wan_pipeline_i2v_2p2.py`): the same skip path adapted for the image-conditioned pipeline (image condition concatenated with the latents, with the required BFHWC↔BCFHW transposes).
- **`generate_wan.py`**: threads `use_magcache` / `magcache_thresh` / `magcache_K` / `retention_ratio` through to both 2.2 pipelines.
- **Configs**:
  - `base_wan_27b.yml` (T2V): MagCache params + official `mag_ratios_base`, and `flow_shift` defaulted to **12.0** (see note below).
  - `base_wan_i2v_27b.yml` (I2V): MagCache params + official I2V-A14B `mag_ratios_base`, with `boundary_ratio=0.900` to align the high→low switch with the curve (`flow_shift` stays at the I2V default of 5.0).
- **Tests** (`wan2_2_magcache_test.py`): host-side validation/schedule/core tests plus a TPU-only end-to-end smoke test.
- **README**: documents MagCache for Wan2.2 T2V and I2V, including the support matrix, config flags, sampling-shift requirement, and benchmark results.

## Important: `flow_shift` alignment

`mag_ratios_base` is calibrated against where the high→low noise boundary lands, which `flow_shift` controls. **Wan2.2 T2V requires `flow_shift=12.0`** (the official A14B sampling shift) — the previous default of `5.0` moved the boundary several steps out of phase, so MagCache skipped at the wrong steps and quality dropped. This PR sets the correct default, which also fixes the off-spec dense baseline. For I2V the official shift is `5.0`, paired with `boundary_ratio=0.900`.

## Results

Measured on a v7x (720×1280, 81 frames, 40 steps), reference = dense (`use_magcache=False`) render with the same seed/config:

| Model | Settings | Speedup | Steps skipped | SSIM | PSNR |
| --- | --- | --- | --- | --- | --- |
| Wan2.2 T2V | `flow_shift=12.0`, `thresh=0.04`, `K=2` | ~1.82× | 18/40 (360s→198s) | ≈0.72 | ≈21.8 dB |
| Wan2.2 I2V | `flow_shift=5.0`, `boundary_ratio=0.900`, `thresh=0.06`, `K=2` | ~1.75× | 17/40 (6.30→3.61 s/step) | ≈0.91 | ≈25.4 dB |

The reference-based metrics mostly reflect *trajectory divergence* — caching nudges the sampler onto a different but equally plausible sample — rather than visible degradation; cached clips are visually hard to tell apart from dense. I2V scores higher because the image conditioning anchors the trajectory. Recalibrating `mag_ratios_base` for a specific dtype/attention kernel can tighten the metric gap further.

## Usage

MagCache is one of several mutually-exclusive caching strategies (CFG Cache, SenCache, MagCache) — enable only one at a time.

```bash
# Wan2.2 T2V
python src/maxdiffusion/generate_wan.py \
  src/maxdiffusion/configs/base_wan_27b.yml \
  use_magcache=True magcache_thresh=0.04 magcache_K=2 ...

# Wan2.2 I2V
python src/maxdiffusion/generate_wan.py \
  src/maxdiffusion/configs/base_wan_i2v_27b.yml \
  use_magcache=True magcache_thresh=0.06 magcache_K=2 ...
```

## Testing

- `wan2_2_magcache_test.py` host-side tests pass (schedule/core logic).
- End-to-end T2V and I2V runs validated on a v7x TPU; speedup and SSIM/PSNR numbers above were collected from those runs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -607,6 +607,18 @@ To generate images, run the following command:
   | --- | --- | --- | --- | --- |
   | **CFG Cache** | `use_cfg_cache: True` | Wan 2.1 T2V, Wan 2.2 T2V/I2V | ~1.2x | FasterCache-style: caches the unconditional branch and applies FFT frequency-domain compensation on skipped steps. |
   | **SenCache** | `use_sen_cache: True` | Wan 2.2 T2V/I2V | ~1.4x | Sensitivity-Aware Caching ([arXiv:2602.24208](https://arxiv.org/abs/2602.24208)): predicts output change via first-order sensitivity S = α_x·‖Δx‖ + α_t·\|Δt\|. Skips the full CFG forward pass when predicted change is below tolerance ε. |
+  | **MagCache** | `use_magcache: True` | Wan 2.1 T2V, Wan 2.2 T2V/I2V | ~1.75–1.9x | [MagCache](https://github.com/Zehong-Ma/MagCache): skips the transformer blocks and reuses the cached block residual when the accumulated magnitude-ratio error stays below `magcache_thresh`, capped at `magcache_K` consecutive skips. Uses a precalibrated per-step `mag_ratios_base` curve, so the skip schedule is deterministic (no data-dependent control flow). |
+
+  For Wan 2.2 (dual-transformer), MagCache uses a single `mag_ratios_base` curve across both phases, forces a full recompute for the first `retention_ratio` fraction of each phase, and resets the cached residual at the high→low boundary. The shipped curves are seeded from the official Wan2.2 values (`base_wan_27b.yml` for T2V, `base_wan_i2v_27b.yml` for I2V); recalibrate for your dtype/attention kernel to tighten the quality gap.
+
+  > **Wan 2.2 T2V requires `flow_shift=12.0`** — it sets where the high→low boundary lands, which is what `mag_ratios_base` is calibrated against. A lower shift (e.g. `5.0`) moves the boundary out of phase, so MagCache skips at the wrong steps and quality drops.
+
+  Benchmarks (7x, A14B, 720×1280, 81 frames, 40 steps, vs dense — SSIM/PSNR largely reflect trajectory divergence, not visible degradation):
+
+  | Variant | Settings | Speedup | SSIM / PSNR |
+  | --- | --- | --- | --- |
+  | T2V | `flow_shift=12.0`, `magcache_thresh=0.04`, `magcache_K=2` | ~1.82× (18/40 skipped) | 0.72 / 21.8 dB |
+  | I2V | `flow_shift=5.0`, `boundary_ratio=0.900`, `magcache_thresh=0.06`, `magcache_K=2` | ~1.75× (17/40 skipped) | 0.91 / 25.4 dB |
 
   To enable a caching mechanism, set the corresponding flag in your config YAML or pass it as a command-line override:
 
@@ -622,7 +634,23 @@ To generate images, run the following command:
     src/maxdiffusion/configs/base_wan_i2v_27b.yml \
     use_cfg_cache=True \
     ...
-  ```
+
+# Example: enable MagCache for Wan 2.2 T2V
+python src/maxdiffusion/generate_wan.py \
+  src/maxdiffusion/configs/base_wan_27b.yml \
+  use_magcache=True \
+  magcache_thresh=0.04 \
+  magcache_K=2 \
+  ...
+
+# Example: enable MagCache for Wan 2.2 I2V
+python src/maxdiffusion/generate_wan.py \
+  src/maxdiffusion/configs/base_wan_i2v_27b.yml \
+  use_magcache=True \
+  magcache_thresh=0.06 \
+  magcache_K=2 \
+  ...
+```
 
 ### Ring Attention
 We added ring attention support for Wan models. Below are the stats for one `720p` (81 frames) video generation (with CFG DP):
--- a/src/maxdiffusion/pipelines/wan/wan_pipeline_2_2.py
+++ b/src/maxdiffusion/pipelines/wan/wan_pipeline_2_2.py
@@ -12,7 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .wan_pipeline import WanPipeline, transformer_forward_pass, transformer_forward_pass_full_cfg, transformer_forward_pass_cfg_cache
+from .wan_pipeline import (
+    WanPipeline,
+    transformer_forward_pass,
+    transformer_forward_pass_full_cfg,
+    transformer_forward_pass_cfg_cache,
+    init_magcache,
+    magcache_step,
+)
 from ...models.wan.transformers.transformer_wan import WanModel
 from typing import List, Union, Optional
 from ...pyconfig import HyperParameters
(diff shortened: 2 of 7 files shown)


src/maxdiffusion/pipelines/wan/wan_pipeline_2_2.py
```diff
@@ -279,6 +305,109 @@ def run_inference_2_2(
...
+
+    # Forced-compute ("retention") zones, in step units: the first
+    # `retention_ratio` fraction of each phase always computes in full. This is
+    # also what flushes the stale residual right after the boundary.
+    high_warmup_end = int(high_noise_steps * retention_ratio)
+    low_warmup_end = high_noise_steps + int((num_inference_steps - high_noise_steps) * retention_ratio)
+
+    cache_count = 0
+    for step in range(num_inference_steps):
+      t = jnp.array(scheduler_state.timesteps, dtype=jnp.int32)[step]
```
Perseus14: We can move `jnp.array(scheduler_state.timesteps, dtype=jnp.int32)` outside and save some compute

```
timesteps = jnp.array(scheduler_state.timesteps, dtype=jnp.int32)
for step in range(num_inference_steps):
    t = timesteps[step]
```

README.md
```diff
@@ -607,6 +607,15 @@ To generate images, run the following command:
...
+
+Wan 2.2 I2V (`base_wan_i2v_27b.yml`) uses the official I2V-A14B `mag_ratios_base` together with the I2V sampling settings (`flow_shift=5.0`, `boundary_ratio=0.900`) that the curve is aligned to. Measured on a v7x (Wan 2.2 A14B I2V, 720×1280, 81 frames, 40 steps, `magcache_thresh=0.06`, `magcache_K=2`): **~1.75× speedup** (17/40 steps skipped, 6.30s → 3.61s per step) at **SSIM ≈ 0.91 / PSNR ≈ 25.4 dB** versus the dense render. Fidelity is higher than T2V because the image conditioning anchors the sampling trajectory, leaving less room for divergence.
```
Perseus14: I think we can reduce the text content here. 

Minor nit: It is 7x and not v7x

src/maxdiffusion/tests/wan/wan2_2_magcache_test.py
```diff
@@ -0,0 +1,389 @@
+"""
```
Perseus14, line 1: Please skip the heavy tests when running in github actions

src/maxdiffusion/tests/wan/wan2_2_magcache_test.py
```diff
@@ -0,0 +1,389 @@
...
+        width=self.config.width,
+        num_frames=self.config.num_frames,
+        num_inference_steps=self.config.num_inference_steps,
+        guidance_scale_low=self.config.guidance_scale_low,
+        guidance_scale_high=self.config.guidance_scale_high,
+        use_magcache=use_magcache,
+    )
+    return videos, time.perf_counter() - t0
+
+  def test_magcache_speedup_and_fidelity(self):
```
Perseus14, line 353: I think it might be better to use logs instead of print statements

Perseus14 (commented):
Thanks for the PR. I have added some comments. PTAL!

Please run a manual linting test.

```
pip install pylint pyink==23.10.0 pytype==2024.2.27
pyink src/maxdiffusion --check --diff --color --pyink-indentation=2 --line-length=125
```

Additionally could you also squash the commits?

src/maxdiffusion/pipelines/wan/wan_pipeline_2_2.py
HadarIngonyama: I agree, but note that for senchache and cfg cache in the same file the array is also inside the loop.

README.md
HadarIngonyama: will do. I will change to 7x, but note that the whole readme has this mistake (and it exists also in other places in the repo)

src/maxdiffusion/tests/wan/wan2_2_magcache_test.py
HadarIngonyama: I think I already have.
The one heavy test — WanMagCacheSmokeTest (loads the 27B model and runs on TPU) — is already guarded with @pytest.mark.skipif(IN_GITHUB_ACTIONS, ...) at line 275, following the same pattern as wan_sen_cache_test.py / wan_cfg_cache_test.py.

The other three classes are intentionally left to run in CI since they're pure-host and don't need a TPU or weights:
- WanMagCacheValidationTest — argument-validation only (__call__ raises on bad config)
- WanMagCacheScheduleTest — host-side skip-schedule math
- WanMagCacheCoreTest — init_magcache / magcache_step unit tests

Model loading is inside the test method (not at import/collection time), so collection stays green in CI. Let me know if you were thinking of something beyond the skipif guard and I'll adjust.

src/maxdiffusion/tests/wan/wan2_2_magcache_test.py
HadarIngonyama: I will fix it, but again, note that this is the case for other tests too - wan_sen_cache_test.py for example.

src/maxdiffusion/pipelines/wan/wan_pipeline_2_2.py
Perseus14: Yes those will be fixed in PR #432

README.md
Perseus14: Thanks, we will put put a PR to fix those

src/maxdiffusion/tests/wan/wan2_2_magcache_test.py
Perseus14: Noted

src/maxdiffusion/tests/wan/wan2_2_magcache_test.py
Perseus14: Looks good!

Perseus14:
LGTM!

I was able to reproduce the results for both T2V and I2V. 

@prishajain1 Could you also take a look?

@HadarIngonyama Please rebase with main. Your current PR includes linting changes to ltx2 lora, which is fixed in main.

src/maxdiffusion/pipelines/wan/wan_pipeline_i2v_2p2.py
```diff
@@ -188,13 +195,31 @@ def __call__(
       use_cfg_cache: bool = False,
       use_sen_cache: bool = False,
       use_kv_cache: bool = False,
+      use_magcache: bool = False,
```
mbohlool, line 198: default values in next 4 lines are inconsistant with wan_pipeline_2_2.py, can you use the same default values please?

src/maxdiffusion/pipelines/wan/wan_pipeline_i2v_2p2.py
```diff
@@ -404,6 +437,117 @@ def run_inference_2_2_i2v(
...
+    # `retention_ratio` fraction of each phase always computes in full. This also
+    # flushes the stale residual right after the boundary.
+    high_warmup_end = int(high_noise_steps * retention_ratio)
+    low_warmup_end = high_noise_steps + int((num_inference_steps - high_noise_steps) * retention_ratio)
+
+    condition_doubled = jnp.concatenate([condition] * 2)
+
+    cache_count = 0
+    for step in range(num_inference_steps):
+      t = jnp.array(scheduler_state.timesteps, dtype=jnp.int32)[step]
```
mbohlool: can you create this outside of the loop and access its index inside of the loop, the same as wan_pipeline_2_2.py?

mbohlool (changes requested):
two minor comments, otherwise looks good.

src/maxdiffusion/pipelines/wan/wan_pipeline_i2v_2p2.py
HadarIngonyama: done

src/maxdiffusion/pipelines/wan/wan_pipeline_i2v_2p2.py
HadarIngonyama: fixed

mbohlool (approved):

Perseus14 (approved):

PR #443, opened 2026-07-10
Cut WAN family time-to-ready 220-304s -> 33-51s: AOT executable cache, zero-exec warmup, converted-weights cache

syhuang22:
# Cut WAN family time-to-ready: AOT executable cache, zero-exec warmup, converted-weights cache

Serving-oriented startup optimizations for the whole WAN family (2.1 / 2.2 T2V, 2.2 I2V, VACE). A process restart previously re-paid tracing, XLA compilation (or persistent-cache lookup), a full torch->flax weight conversion, and two real denoise steps of warmup. All four are either memoized on disk or eliminated.

## Mechanisms

1. **Per-shape AOT executable cache** (`aot_cache.py`): `cached_jit` replaces `jax.jit` at definition sites (identical behavior until `install()`). After warmup the compiled executables are serialized (`jax.experimental.serialize_executable`); later processes deserialize on background threads — no trace, no lowering, no cache lookup. One file per input signature, so resolutions never collide. Wired into the transformer forward passes (base, CFG, cfg-cache, VACE) and VAE encode/decode.
2. **Zero-exec warmup** (`aot_cache.warmup_mode()`): compilation only needs avals. During warmup, wrapped calls lower+compile (or reuse the deserialized executable) and return sharded zeros, so downstream executables (scheduler, VAE) still warm faithfully — without paying real 14B denoise steps.
3. **Converted-weights cache** (`wan_utils.py`, `converted_weights_dir`): the torch->flax conversion (per-tensor transpose + scan-stack + dtype cast) is a pure function of the checkpoint; it is now saved once as per-tensor `.npy` + manifest and re-loaded with parallel reads. Self-validating against eval shapes and the cast policy (mismatch = silent reconvert).

Both caches are off by default (`aot_cache_dir: ''`, `converted_weights_dir: ''`).

## Results (v7x-8, 720p/81f, 2D-ring recipe, same warm JAX persistent cache on both sides; only the two cache knobs differ)

| Pipeline | Load (off -> hit) | Compile window (off -> hit) | Time-to-ready (off -> hit) |
|---|---|---|---|
| WAN 2.1 T2V 14B | 24.2s -> 25.0s | 195.5s -> **8.3s** (23.5x) | 219.7s -> **33.3s** |
| WAN 2.2 T2V A14B | 29.6s -> 28.1s | 55.9s -> **8.7s** (6.4x) | 85.5s -> **36.8s** |
| WAN 2.2 I2V A14B | 29.2s -> 27.9s | 151.7s -> **16.3s** (9.3x) | 180.9s -> **44.2s** |
| WAN 2.1 VACE 14B | 29.7s -> 29.6s | 274.6s -> **21.5s** (12.8x) | 304.3s -> **51.1s** |

First run per (model, shape) builds the caches and lands in between (compile 11-33s).

Highlights: VAE decode cold compile 144.9s -> 0.2s (2.1), I2V image VAE encode 89.5s -> 1.0s, VACE (unscanned, so all 40 blocks were traced every process) 274.6s -> 21.5s.

Load is flat in this A/B because both sides already include this PR's loader changes (fused cast, chunked staged put, threaded 2.1 component load) and the remaining critical path is the CPU text-encoder load; measured against the previous loader the transformer host path alone is 26.2s -> 8.5s, verified bitwise-equal on all 1095 params.

## Correctness

- Denoise step time identical across all variants (5.9s +-0.4 run noise); videos generated with all mechanisms on validated frame-exact (h264, 1280x720, 81 frames, zero decode errors on full decode) for all four pipelines.
- Loader-path equivalence verified bitwise (1095/1095 params) between the fused-cast path and the historical two-pass path.
- The AOT cache is keyed by a config fingerprint (model, attention, flash block sizes, mesh, dtypes, scan, jax version) plus a per-call input signature; unknown signatures fall back to plain jit and are serialized after the next warmup. Executables deserialize with mesh-ordered `execution_devices`.
- Unit tests: `aot_cache_test.py` (9, incl. cross-process signature determinism and in-trace inlining) and `converted_weights_cache_test.py` (5, incl. bf16 bit-view round-trip and cache invalidation).

## Example

`end_to_end/tpu/run_wan_fast_inference.sh` — end-to-end serving example (tuned v7 flag set + both caches): `./run_wan_fast_inference.sh 22 40 "a cat surfing"`.

### DIFF
--- a/end_to_end/tpu/run_wan_fast_inference.sh
+++ b/end_to_end/tpu/run_wan_fast_inference.sh
@@ -1,0 +1,90 @@
+#!/bin/bash
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# WAN T2V fast-serving example: AOT executable cache + converted-weights
+# cache + zero-exec warmup, with a tuned v7 2D-ring attention recipe.
+#
+# First run per (model, shape) pays one-time conversion + compile and
+# populates the caches; every later process start is ~25s to ready.
+#
+# Usage:
+#   ./run_wan_fast_inference.sh [21|22] [steps] ["prompt..."]
+# Env overrides:
+#   WAN_CACHE_ROOT   cache root (default ~/.cache/maxdiffusion_wan)
+#   OUTPUT_DIR       video/metrics output (default /tmp/wan_out)
+#   COMPILE_TE=true  torch.compile the text encoder (adds ~30s to load,
+#                    saves ~10s/encode; worth it for long-lived processes)
+set -u
+MODEL=${1:-22}
+STEPS=${2:-40}
+PROMPT=${3:-""}
+
+PROJECT_ROOT="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/../.." &> /dev/null && pwd)"
+cd "$PROJECT_ROOT" || exit 1
+export PYTHONPATH="$PROJECT_ROOT/src:${PYTHONPATH:-}"
+export HF_HUB_ENABLE_HF_TRANSFER=1
+export JAX_DEFAULT_MATMUL_PRECISION=bfloat16
+export TORCHINDUCTOR_FX_GRAPH_CACHE=1
+
+CACHE_ROOT=${WAN_CACHE_ROOT:-$HOME/.cache/maxdiffusion_wan}
+OUTPUT_DIR=${OUTPUT_DIR:-/tmp/wan_out}
+mkdir -p "$CACHE_ROOT/jax" "$CACHE_ROOT/aot_wan$MODEL" "$CACHE_ROOT/converted" "$OUTPUT_DIR"
+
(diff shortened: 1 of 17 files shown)


end_to_end/tpu/run_wan_fast_inference.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
```
eltsai, line 1: I am not sure if we want to upload the inference script here, will wait for @entrpn 's input

entrpn: I think its ok to upload it, but lets also reference it in the readme.

entrpn (dismissed):

eltsai (dismissed):
LGTM! I was able to reproduce the stats on v7x

src/maxdiffusion/aot_cache.py
```diff
@@ -0,0 +1,413 @@
...
+import jax
+import jax.numpy as jnp
+from jax.experimental import serialize_executable
+
+from maxdiffusion import max_logging
+
+_FORMAT_VERSION = 1
+
+
+def _dynamic_signature(args: tuple, kwargs: dict) -> str:
```
csgoogle, line 75: Thanks for the changes, these are really helpful.

Wondering if we can use orbax for serialization and deserialization. I've made a change https://github.com/AI-Hypercomputer/maxdiffusion/pull/406, that reduces the load time to <60s, however it doesn't touch the compile time. Do you think if it's possible to reuse it.

syhuang22: The compile side is orthogonal. Executables aren't arrays, so orbax can't cache them — that's what jax.experimental.serialize_executable covers here, and it's the bigger share of the win.

ninatu (dismissed):

eltsai (approved):

entrpn (approved):

PR #444, opened 2026-07-12
Add `heads_per_tile` (multi-head-per-tile) support to Ulysses ring attention

edgexyz:
## Summary

This PR adds a `heads_per_tile` (**mhpt** — multi-head-per-tile) option to the tokamax / Ulysses **ring** splash-attention forward kernel (`_splash_attention_forward_ring_raw`). Instead of processing a single attention head per Pallas grid program, the kernel can now process `heads_per_tile` heads per program, amortizing per-tile overhead (grid launch, mask/scalar prefetch, VMEM pipelining) across several heads. This mirrors the existing mhpt fast-path that already exists for the non-ring custom splash kernel (`ulysses_custom`), extending the same optimization to the `ulysses_ring` / `tokamax_ring` attention modes used by Wan2.2 I2V on TPU v6e.

`heads_per_tile` is a **pure tiling/scheduling choice**: for a fixed input it must produce numerically identical results to `heads_per_tile=1`. The correctness evidence below shows it does (bit-identical output), and the perf motivation is that it lets a v6e-16 run tile 40 DiT heads into 20 (hpt=2) grid programs.

- **Branch:** `ulysses-ring-vmem-passthrough` — base `main` @ `f62927f`
- **Commits (vs `main`):**
  - `049210a` — `feat(splash-attention): heads_per_tile + configurable vmem_limit_bytes for ulysses ring attention`. The feature itself; the block-index detail below is already correct here, and the mhpt invariance unit test ships in this same commit.
  - `b47433e` — `docs(splash-ring): document mhpt kernel + dispatch design rationale`.
  - `0d5d0f8` — `test(splash-ring): make mhpt heads_per_tile test pass under multi-host`.
- **Files changed (branch vs `main`):** 5 files, `+372 / −25` (feature commit `049210a` alone: 4 files, `+311 / −25`).

## Motivation

The Wan2.2 DiT (`num_attention_heads = 40`) runs self/cross attention through splash attention. On the ring path each head was launched as its own grid program along grid dim 0. Batching multiple heads per program reduces launch/prefetch overhead and improves VMEM reuse. The non-ring custom kernel already had an mhpt path; this PR brings feature parity to the ring kernel so `attention=ulysses_ring flash_block_sizes={... "heads_per_tile":N}` is usable.

## What changed

| File | Change |
| --- | --- |
| `src/maxdiffusion/kernels/splash_attention/splash_attention_kernel.py` | New `flash_attention_kernel_mhpt` (per-tile inner loop over `heads_per_tile` heads with per-head `m`/`l`/`o` scratch); `SplashConfig.heads_per_tile` field + validation; `_splash_attention_forward_ring_raw` now selects the mhpt kernel, sizes the head-dim `BlockSpec` block to `heads_per_tile`, adjusts the grid to `num_q_heads // heads_per_tile`, and appends `_hpt<N>` to the kernel name. |
| `src/maxdiffusion/max_utils.py` | New `TokamaxRingFlashBlockSizes` hashable carrier and a branch in `get_flash_block_sizes` so `ulysses_ring` / `tokamax_ring` propagate `heads_per_tile` (plus block sizes) from `config.flash_block_sizes`. |
| `src/maxdiffusion/models/attention_flax.py` | `convert_to_tokamax_splash_config` now threads `heads_per_tile` (default 1) into the `SplashConfig`. |

The three files above are the core mechanism. The rest of the branch: the mhpt invariance unit test in `ring_attention_kernel_test.py` (added in feature commit `049210a`) and the `assert_allclose_mcjax` helper in `splash_attention_test_utils.py` (`0d5d0f8`) — see *Correctness evidence*; the in-code kernel/dispatch rationale docstring (`b47433e`); and a configurable `vmem_limit_bytes` bundled into the feature commit (out of scope for this summary).

### Feature scope / guards
The mhpt ring path deliberately raises `NotImplementedError` outside the validated regime, so misconfigurations fail rather than silently miscomputing:
- static ring grids only (no dynamic/active-block grids),
- full **MHA only** (`num_q_heads == num_kv_heads`; MQA/GQA rejected),
- static **FullMask** only (no block/partial masks, no `q_sequence`, no `mask_function`),
- no attention sinks, no max-logit bounds, no logits soft-cap,
- `HEAD_DIM_MINOR` Q/K/V layouts only.

`heads_per_tile=1` is the default everywhere and is completely unaffected (the mhpt kernel and block-size changes are only engaged when `heads_per_tile > 1`).

## Key correctness detail — Pallas block-index vs. element-index

The subtlest part of `heads_per_tile>1` is a **Pallas block-index vs. element-index** trap: getting it wrong produces broken output while still exiting cleanly (`exit_status=0`), so it is documented here as the crux of the correctness story. The kernel in this PR gets it right; the explanation below is the rationale (and a guard against reintroducing it).

In a Pallas `BlockSpec`, the index map returns a **block index**; Pallas multiplies it by the block size to get the element offset. The head-dim block size is `heads_per_tile`, so the index maps must **not** also multiply the program id — `h` is already the head-*tile* index:

```python
# WRONG: head_index(h) = h * heads_per_tile used as the block index
# → element offset = (h * heads_per_tile) * heads_per_tile = h * heads_per_tile**2
q_index_map   = unravel(lambda h, i, j: from_head_minor((head_index(h), i, 0), q_layout))
out_index_map = unravel(lambda h, i, j: (head_index(h), i, 0))
prefix        = (_div(head_index(h), q_heads_per_kv_head),)   # k / v

# RIGHT: h is already the head-*tile* index; the block size is heads_per_tile,
# so the block index is simply h → element offset = h * heads_per_tile
q_index_map   = unravel(lambda h, i, j: from_head_minor((h, i, 0), q_layout))
out_index_map = unravel(lambda h, i, j: (h, i, 0))
prefix        = (_div(h, q_heads_per_kv_head),)               # k / v
```

**Failure mode if done wrong** (Wan DiT, `num_q_heads=40`, `heads_per_tile=2`, grid dim 0 = `40/2 = 20` programs): program `h` would write output heads starting at `h*2*2 = 4h` instead of `2h`. Only heads `{0,1,4,5,8,9,…,36,37}` plus the clamped tail `{38,39}` would ever be written — **18 of 40 heads would stay at their uninitialized/zero value**, with reads scrambled the same way. The pipeline still runs to completion (`exit_status=0`) but the decoded video is effectively static — which is why this is called out explicitly.

The index maps align with the **already-correct reference mhpt kernel** in `src/maxdiffusion/kernels/custom_splash_attention.py` (the `ulysses_custom` path), whose `q_index_map` returns `(h, i, 0)`. The `q_heads_per_kv_head != 1` guard matches that file's `assert num_q_heads == num_kv_heads` (mhpt requires no GQA).

## Correctness evidence

Verified on a live **TPU v6e-16** (`ici_context_parallelism=4 ici_tensor_parallelism=4`, `ulysses_shards=4`), Wan2.2-I2V-A14B, running the *same* generate config three times and changing only the attention setting / code:

| Run | `attention` | `heads_per_tile` | output `md5sum` | mp4 size | 4-worker exit |
| --- | --- | --- | --- | --- | --- |
| Baseline | `ulysses_ring` | **1** | `c8a0ff4182551867cb0bb75c768c6d72` | 496,148 B | all `0` |
| **This PR** | `ulysses_ring` | **2** | **`c8a0ff4182551867cb0bb75c768c6d72`** | 496,148 B | all `0` |
| Wrong block-index (contrast) | `ulysses_ring` | 2 | `b31f17bf126bfc41844b092e176c741e` | 5,128 B | all `0` |

**The `heads_per_tile=2` output is byte-for-byte identical (same md5) to the `heads_per_tile=1` baseline.** Because mhpt is a pure tiling optimization, bit-identical output against the single-head baseline is the strongest correctness signal available — it proves the tiling changes scheduling only, not results. The third row is a development-time contrast: with the wrong block-index (see *Key correctness detail*) the run still exits `0` but produces a differing md5 and a 5 KB, ≈40 B/frame (i.e. static) video — the failure mode the exit code alone hid.

Additional cross-checks performed during the investigation:
- A `mask_padding_tokens=False` variant of the wrong-index run stayed 5 KB, ruling out segment-id / padding masking as the cause and isolating it to head addressing.
- The XLA trace kernel name carried the expected `_hpt2` suffix, confirming the run actually exercised the ring mhpt kernel (not the `ulysses_custom` path).

### Automated unit test — `heads_per_tile` invariance (multi-host)

`RingAttentionHeadsPerTileTest.test_heads_per_tile_matches_single_head` (`src/maxdiffusion/kernels/splash_attention/ring_attention_kernel_test.py`) locks in the invariance this PR relies on: for fixed random MHA inputs it runs the ring kernel with `heads_per_tile ∈ {2, 4}` and asserts the output matches the `heads_per_tile=1` baseline (`rtol = atol = 5e-3`; `ring_size=2`, `num_heads=4`, `seq_len=2048`, `HEAD_DIM=128`, bf16, static `FullMask`).

Run it — on **all** workers at once (the v6e-16 slice only initializes when every host starts together; a lone host hangs in `jax.devices()`):

```bash
# fill in for your slice:
TPU_NAME=<your v6e-16 node>   ZONE=<its zone>
MAXDIFFUSION_HOME=<path to maxdiffusion checkout>   PYTHON=<tpu venv python>

gcloud compute tpus tpu-vm ssh "$TPU_NAME" --zone="$ZONE" --worker=all \
  --command="cd $MAXDIFFUSION_HOME && \
    $PYTHON -m pytest \
      src/maxdiffusion/kernels/splash_attention/ring_attention_kernel_test.py \
      -k test_heads_per_tile_matches_single_head -q"
# each worker prints: "2 passed, 16 deselected"  (exit 0)
```

Verified on a live **TPU v6e-16**, launched on all 4 workers simultaneously (the slice cannot initialize on a single host):

| host (process_index 0–3) | result |
| --- | --- |
| all 4 hosts | `2 passed, 16 deselected` · exit 0 · ≈11 s each |

Both parameter cases (`hpt=2`, `hpt=4`) pass on **every** host. Making the test green under multi-controller (multi-host) JAX required one fix (commit `0d5d0f8` on `ulysses-ring-vmem-passthrough`): the test builds its mesh from `jax.devices()[:ring_size]` (all on process 0), so the stock `np.testing.assert_allclose` — which pulls the sharded `jax.Array` back to host — succeeded only on the owning process and raised `RuntimeError: ... spans non-addressable devices` on the other three, failing the test on all hosts but one. A new `SplashAttentionTestCase.assert_allclose_mcjax` helper evaluates the comparison on-device, reads the scalar verdict only on the owner, and broadcasts it to every process via `multihost_utils.broadcast_one_to_all` (one matching collective per process → no deadlock). The test therefore passes on all hosts while still genuinely running the kernel on TPU — no `skipTest`.

### Reproduce (end-to-end video correctness)

Wan2.2-I2V-A14B `generate_wan.py` on a v6e-16, run **once with `heads_per_tile=1`
(baseline) and once with `=2`**, everything else identical, then compare the produced
mp4s. Launch on `--worker=all` so all 4 hosts start together (single-host hangs):

```bash
# fill in for your slice:
TPU_NAME=<your v6e-16 node>   ZONE=<its zone>
MAXDIFFUSION_HOME=<path to maxdiffusion checkout>   PYTHON=<tpu venv python>   OUT=<output dir>

CFG=src/maxdiffusion/configs/base_wan_i2v_27b.yml
COMMON="height=576 width=1024 num_inference_steps=1 per_device_batch_size=0.0625 \
guidance_scale_low=1.0 guidance_scale_high=1.0 \
ici_context_parallelism=4 ici_tensor_parallelism=4"
BLK() { echo "{\"block_q\":1024,\"block_kv\":1024,\"block_kv_compute\":256,\"block_q_dkv\":1024,\"block_kv_dkv\":1024,\"block_kv_dkv_compute\":256,\"block_q_dq\":1024,\"block_kv_dq\":1024,\"heads_per_tile\":$1}"; }

for N in 1 2; do
  gcloud compute tpus tpu-vm ssh "$TPU_NAME" --zone="$ZONE" --worker=all \
    --command="cd $MAXDIFFUSION_HOME && \
      $PYTHON src/maxdiffusion/generate_wan.py $CFG $COMMON \
        attention=ulysses_ring flash_block_sizes='$(BLK $N)' \
        run_name=hpt$N output_dir=$OUT/hpt$N"
done

# worker-0's mp4 from each run must be byte-identical:
md5sum "$OUT"/hpt1/**/wan_output_0_0.mp4 "$OUT"/hpt2/**/wan_output_0_0.mp4
```

Smoke config used for the evidence above: `576x1024`, `num_inference_steps=1`,
`per_device_batch_size=0.0625`, `guidance_scale_low=guidance_scale_high=1.0`, the
`flash_block_sizes` shown (with `heads_per_tile` = 1 or 2). Cold compile 213–423 s, warm
generation ≈15.6 s/step, all 4 workers `exit_status=0`. (Internally these runs were driven
by the `run_generate_wan_profiler.sh` launcher, which wraps exactly the `generate_wan.py`
invocation above and fans it out to every worker.)

## Scope, limitations & follow-ups

- **Forward only.** This is the inference/generate forward kernel; the mhpt fast-path is not wired into the backward kernel (fine for generation, which is all `ulysses_ring` serves here).
- **MHA + static FullMask only** (enforced by the guards above).
- **Unit test (mhpt invariance).** `test_heads_per_tile_matches_single_head` asserts `mhpt(N) == mhpt(1)` for `N ∈ {2, 4}`, guarding against exactly the block-index class of error in *Key correctness detail*. It ships in the feature commit (`049210a`) and was made multi-host-safe in `0d5d0f8` (the new `assert_allclose_mcjax` helper); it passes on a live TPU v6e-16 across all 4 hosts (see *Automated unit test*) — i.e. it runs on the real cluster, not just `interpret=True`.

### DIFF
--- a/src/maxdiffusion/kernels/splash_attention/ring_attention_kernel_test.py
+++ b/src/maxdiffusion/kernels/splash_attention/ring_attention_kernel_test.py
@@ -21,6 +21,7 @@ from absl.testing import absltest
 from absl.testing import parameterized
 import jax
 from jax import random
+from jax.experimental import multihost_utils
 import jax.numpy as jnp
 import numpy as np
 from . import base
@@ -158,6 +159,138 @@ class RingAttentionTest(test_utils.SplashAttentionTestCase):
       self._assert_allclose(dq, dq_ref, rtol=1e-2, atol=1e-2)
       self._assert_allclose(dk, dk_ref, rtol=1e-2, atol=1e-2)
       self._assert_allclose(dv, dv_ref, rtol=1e-2, atol=1e-2)
+
+
+class RingAttentionHeadsPerTileTest(test_utils.SplashAttentionTestCase):
+  """`heads_per_tile` (multi-head-per-tile) invariance for ring attention.
+
+  heads_per_tile is a pure tiling/scheduling choice for the forward kernel: for a
+  fixed input, running with heads_per_tile=N must produce the same output as
+  heads_per_tile=1. This guards against the block-index class of bug (a wrong
+  head-tile mapping compiles and runs but silently returns garbage).
+  """
+
+  def setUp(self):
+    if jax.default_backend() != "tpu":
+      self.skipTest("Multi-head-per-tile ring attention runs on TPU.")
+    super().setUp()
+
+  @parameterized.product(
+      heads_per_tile=[2, 4],
+      head_dim=[128],
+      dtype=[jnp.bfloat16],
+  )
+  def test_heads_per_tile_matches_single_head(self, heads_per_tile, head_dim, dtype):
+    # Use ALL devices so the sharded outputs are addressable on every process
+    # (a subset mesh like jax.devices()[:2] lives only on process 0, making the
+    # result non-addressable on the other hosts). jax.devices() is itself a
+    # global barrier, so every host must launch this test together regardless.
+    ring_size = jax.device_count()
+    num_heads = 8  # MHA (num_q_heads == num_kv_heads); divisible by heads_per_tile.
+    if ring_size < 2:
+      self.skipTest(f"This test needs at least 2 devices, but has {ring_size}.")
+
+    ring_axis = "ring"
+    devices = np.asarray(jax.devices()).reshape(1, ring_size)
+    mesh = jax.sharding.Mesh(devices, ("heads", ring_axis))
+    seq_len = 1024 * ring_size
+
+    k1, k2, k3 = random.split(random.key(0), 3)
+    scale = head_dim**-0.5
+    q = random.normal(k1, (num_heads, seq_len, head_dim), dtype=dtype) * scale
+    k = random.normal(k2, (num_heads, seq_len, head_dim), dtype=dtype) * scale
+    v = random.normal(k3, (num_heads, seq_len, head_dim), dtype=dtype) * scale
+
+    # The mhpt fast path supports full MHA + static FullMask + HEAD_DIM_MINOR only.
+    mask = mask_lib.FullMask(_shape=(seq_len, seq_len))
+    q_spec = P(None, ring_axis, None)
+    kv_spec = q_spec
+
+    def run_multi_head_kernel(hpt):
+      config = splash.SplashConfig.get_default()
+      config = dataclasses.replace(
+          config,
+          use_base2_exp=False,
+          fuse_reciprocal=True,
+          heads_per_tile=hpt,
+      )
+      ring_kernel = ring_attention_kernel.make_ring_attention(
+          mask,
+          is_mqa=False,
+          ring_axis=ring_axis,
+          config=config,
+          save_residuals=False,
+          q_seq_shards=ring_size,
+          kv_seq_shards=ring_size,
+      )
+      kernel_spec = ring_kernel.manual_sharding_spec()
+
+      @partial(
+          jax.shard_map,
+          mesh=mesh,
+          in_specs=(kernel_spec, q_spec, kv_spec, kv_spec, None),
+          out_specs=q_spec,
+          check_vma=False,
+      )
+      def ring_attn(ring_kernel, q, k, v, segment_ids):
+        return ring_kernel(q, k, v, segment_ids)
+
+      return ring_attn(ring_kernel, q, k, v, None)
+
+    out_ref = run_multi_head_kernel(1)  # baseline: single head per tile (flash_attention_kernel)
+    out_mhpt = run_multi_head_kernel(heads_per_tile)  # multi-head-per-tile (flash_attention_kernel_mhpt)
+
+    # Pure tiling => numerically equivalent to the single-head-per-tile baseline.
+    # Outputs are sharded across all hosts; all-gather to a fully-replicated host
+    # array on every process, then compare with the standard helper.
+    out_mhpt = multihost_utils.process_allgather(out_mhpt, tiled=True)
+    out_ref = multihost_utils.process_allgather(out_ref, tiled=True)
+    self._assert_allclose(out_mhpt, out_ref, rtol=5e-3, atol=5e-3)
+
+
+class RingAttentionHeadsPerTileGuardTest(test_utils.SplashAttentionTestCase):
+  """Negative scenarios: configurations the heads_per_tile > 1 fast path rejects.
+
+  The mhpt kernel only supports full-MHA static-FullMask ring attention;
+  `_validate_heads_per_tile_support` must reject everything else with
+  NotImplementedError instead of letting the kernel silently miscompute.
+  Pure-Python validation, so no TPU is required.
+  """
+
+  _EMPTY_MASK_INFO = splash.MaskInfo(None, None, None, None, None, None, None)
+
+  def _validate(self, **overrides):
+    kwargs = dict(
+        config=splash.SplashConfig.get_default(),
+        dynamic_grid=False,
+        is_mqa=False,
+        q_heads_per_kv_head=1,
+        mask_info=self._EMPTY_MASK_INFO,
+        mask_function=None,
+        sinks=None,
+        max_logit_value=None,
+    )
+    kwargs.update(overrides)
+    splash._validate_heads_per_tile_support(**kwargs)  # pylint: disable=protected-access
+
+  def test_supported_config_is_accepted(self):
+    self._validate()  # full MHA + static FullMask: must not raise
+
+  def test_mqa_is_rejected(self):
+    with self.assertRaisesRegex(NotImplementedError, "MHA"):
+      self._validate(is_mqa=True)
+
+  def test_gqa_is_rejected(self):
+    with self.assertRaisesRegex(NotImplementedError, "MHA"):
+      self._validate(q_heads_per_kv_head=2)
+
+  def test_dynamic_grid_is_rejected(self):
+    with self.assertRaisesRegex(NotImplementedError, "static ring attention grids"):
+      self._validate(dynamic_grid=True)
+
+  def test_non_full_mask_is_rejected(self):
+    with self.assertRaisesRegex(NotImplementedError, "FullMask"):
+      self._validate(mask_function=lambda *args: True)
 
 
 if __name__ == "__main__":
--- a/src/maxdiffusion/kernels/splash_attention/ring_attention_kernel_test.py
+++ b/src/maxdiffusion/kernels/splash_attention/ring_attention_kernel_test.py
@@ -21,6 +21,7 @@ from absl.testing import absltest
 from absl.testing import parameterized
 import jax
 from jax import random
+from jax.experimental import multihost_utils
 import jax.numpy as jnp
 import numpy as np
 from . import base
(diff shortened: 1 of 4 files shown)


Perseus14:
@eltsai @csgoogle Could you please review this?

src/maxdiffusion/max_utils.py
```diff
@@ -633,11 +633,29 @@ class CustomFlashBlockSizes:
   vmem_limit_bytes: int | None = None
 
 
+@dataclasses.dataclass(frozen=True)
+class TokamaxRingFlashBlockSizes:
```
csgoogle: Do you think it's better to rename the CustomFlashBlockSizes and reuse that for all code paths

src/maxdiffusion/kernels/splash_attention/splash_attention_test_utils.py
```diff
@@ -76,6 +77,43 @@ def _assert_allclose(self, x, y, **kwargs):
...
+    fully-sharded (every-process-owns-a-shard) arrays use `_assert_allclose`.
+    """
+    if x.dtype == np.dtype(jnp.bfloat16):
+      x = x.astype(jnp.float32)
+    if y.dtype == np.dtype(jnp.bfloat16):
+      y = y.astype(jnp.float32)
+    self.assertTupleEqual(x.shape, y.shape)
+    ok = jnp.all(jnp.abs(x - y) <= atol + rtol * jnp.abs(y))
+    max_err = jnp.max(jnp.abs(x - y))
+    is_owner = jax.process_index() == 0
```
csgoogle: Isn't it better to do all gather and verify it?

src/maxdiffusion/max_utils.py
edgexyz: Both carriers existed for the same reason — the upstream jax splash_attention_kernel.BlockSizes is frozen + slotted and can't hold the extra fields (block_kv_compute_in for custom; block_*_dkv / heads_per_tile / vmem_limit_bytes for ring), and a plain dict isn't hashable.

Merged TokamaxRingFlashBlockSizes into CustomFlashBlockSizes (field union) and dropped the separate class; the ring path now returns CustomFlashBlockSizes too. Both consumers read by field name via getattr, so no consumer changes were needed — only the two returns in get_flash_block_sizes, which keep their own populate logic (custom uses lenient .get(...); ring requires the dkv keys and forces use_fused_bwd_kernel=True). Done in 0cd23dd.

src/maxdiffusion/kernels/splash_attention/splash_attention_test_utils.py
edgexyz: The test now builds the mesh from all devices (ring_size = jax.device_count()), so the outputs are sharded across every process; both results are then gathered with multihost_utils.process_allgather(..., tiled=True) and compared with the standard assert_allclose. The bespoke assert_allclose_mcjax helper is deleted.

Verified on a live v6e-16 (4 hosts x 4 chips, launched with --worker=all): all 4 hosts report 2 passed (hpt=2 and hpt=4 vs the hpt=1 baseline) in ~34s each. Done in 758cb1f.

csgoogle:
LGTM, please squish all the commits

csgoogle (dismissed):

src/maxdiffusion/kernels/splash_attention/splash_attention_kernel.py
```diff
@@ -940,6 +1076,37 @@ def _splash_attention_forward_ring_raw(
   q_layout = config.q_layout
   k_layout = config.k_layout
   v_layout = config.v_layout
+  use_heads_per_tile = heads_per_tile > 1
```
csgoogle, line 1124: to make code more readable, you can move this to a method

src/maxdiffusion/kernels/splash_attention/ring_attention_kernel_test.py
```diff
@@ -160,5 +161,92 @@ def ring_attn(ring_kernel, q, k, v, segment_ids):
...
+    if jax.default_backend() != "tpu":
+      self.skipTest("Multi-head-per-tile ring attention runs on TPU.")
+    super().setUp()
+
+  @parameterized.product(
+      heads_per_tile=[2, 4],
+      head_dim=[128],
+      dtype=[jnp.bfloat16],
+  )
+  def test_heads_per_tile_matches_single_head(self, heads_per_tile, head_dim, dtype):
```
csgoogle, line 183: To make tests more robust you can also add tests cases for negative scenarios like MQA, GQA which you mentioned. Also minor nit, keep names more descriptive, like instead of `run` use may be `run_multi_head_kernel`

csgoogle (dismissed):

csgoogle (approved):

ThomasNing:
@Perseus14  Hi Rishabh, could you take another look?

src/maxdiffusion/max_utils.py
```diff
@@ -651,6 +660,20 @@ def get_flash_block_sizes(config):
           heads_per_tile=user_block_sizes.get("heads_per_tile"),
           vmem_limit_bytes=user_block_sizes.get("vmem_limit_bytes"),
       )
+    if attention_uses_tokamax_ring and "heads_per_tile" in user_block_sizes:
+      return CustomFlashBlockSizes(
+          block_q=user_block_sizes.get("block_q_dkv", user_block_sizes["block_kv"]),
+          block_kv=user_block_sizes["block_kv"],
+          block_kv_compute=user_block_sizes["block_kv_compute"],
+          block_q_dkv=user_block_sizes["block_q_dkv"],
```
mbohlool: 🟡 **Important**: Direct dictionary indexing `user_block_sizes["block_q_dkv"]`, `user_block_sizes["block_kv_dkv"]`, and `user_block_sizes["block_kv_dkv_compute"]` will raise `KeyError` if a user provides a forward-only inference config (e.g. `{"block_q": 1024, "block_kv": 1024, "block_kv_compute": 256, "heads_per_tile": 2}`).

Consider using `.get(...)` with fallbacks matching the non-MHPT Tokamax path below:
```python
if attention_uses_tokamax_ring and "heads_per_tile" in user_block_sizes:
  block_q = user_block_sizes.get("block_q_dkv", user_block_sizes.get("block_q", user_block_sizes.get("block_kv")))
  block_kv = user_block_sizes.get("block_kv")
  block_kv_compute = user_block_sizes.get("block_kv_compute")
  return CustomFlashBlockSizes(
      block_q=block_q,
      block_kv=block_kv,
      block_kv_compute=block_kv_compute,
      block_q_dkv=user_block_sizes.get("block_q_dkv", block_q),
      block_kv_dkv=user_block_sizes.get("block_kv_dkv", block_kv),
      block_kv_dkv_compute=user_block_sizes.get("block_kv_dkv_compute", block_kv_compute),
      block_q_dq=None,
      block_kv_dq=None,
      use_fused_bwd_kernel=True,
      heads_per_tile=user_block_sizes.get("heads_per_tile"),
      vmem_limit_bytes=user_block_sizes.get("vmem_limit_bytes"),
  )
```

src/maxdiffusion/kernels/splash_attention/splash_attention_kernel.py
```diff
@@ -482,6 +489,131 @@ def end():
...
+
+  @pl.when(should_initialize)
+  def init():
+    o_scratch_ref[...] = jnp.zeros_like(o_scratch_ref)
+    m_scratch_ref[...] = jnp.full_like(m_scratch_ref, mask_value)
+    l_scratch_ref[...] = jnp.zeros_like(l_scratch_ref)
+
+  def body(kv_compute_index, _):
+    slice_k = pl.ds(kv_compute_index * bkv_compute, bkv_compute)
+    bkv_repeats, rem = divmod(bkv_compute, NUM_LANES)
```
mbohlool: 💡 **Suggestion**: Since `bkv_compute` and `NUM_LANES` are compile-time constants, `bkv_repeats, rem = divmod(bkv_compute, NUM_LANES)` can be hoisted outside `body` alongside `head_dim_v_repeats = pl.cdiv(head_dim_v, NUM_LANES)` to avoid redundant recomputations during tracing.

src/maxdiffusion/kernels/splash_attention/splash_attention_kernel.py
```diff
@@ -482,6 +489,131 @@ def end():
...
+  This is a lean, dedicated kernel rather than a branch inside `flash_attention_kernel`
+  on purpose: mhpt supports only a narrow regime (full MHA, static FullMask,
+  HEAD_DIM_MINOR, no sinks / max-logit / soft-cap), so keeping it separate avoids adding
+  a head loop crossed with all of `flash_attention_kernel`'s branches to the hot path.
+  Mirrors `custom_splash_attention._flash_attention_kernel_mhpt` (the non-ring mhpt path).
+  Supported regime is enforced by guards in `_splash_attention_forward_ring_raw`.
+  """
+  del active_rows_ref, active_cols_ref, mask_next_ref
+  del bounds_start_ref, bounds_end_ref, block_mask_ref
+  del sinks_ref, mask_ref, q_sequence_ref, max_logit_value_ref
```
mbohlool: 🟢 **Nit**: `logsumexp_ref` is present in the `flash_attention_kernel_mhpt` argument list but is neither used nor deleted (`del logsumexp_ref`), whereas all other unused input/output references are explicitly deleted here. Consider adding `del logsumexp_ref` (or `assert logsumexp_ref is None`) for consistency.

PR #446, opened 2026-07-14
feat: Add Z-Image and Z-Image-Turbo support to MaxDiffusion

csgoogle:
# Add Z-Image and Z-Image-Turbo support

Adds the Tongyi-MAI Z-Image family to MaxDiffusion: a 6.15B flow-matching DiT, its Flux-format VAE, and a Qwen3-4B text encoder — all running in JAX on TPU, with an Orbax cache for the whole pipeline.

## What's here

| | |
|---|---|
| `models/z_image/transformer_z_image.py` | NNX port of the 6.15B denoiser |
| `models/z_image/z_image_utils.py` | Streaming Diffusers→NNX checkpoint conversion |
| `pipelines/z_image/z_image_pipeline.py` | Runtime pipeline: encode → denoise → decode |
| `checkpointing/z_image_checkpointer.py` | All loading, plus the Orbax cache |
| `models/qwen3_utils.py` | Qwen3 checkpoint conversion, split out of `qwen3_flax.py` |
| `generate_zimage.py` | Driver, structured like `generate_wan.py` |
| `configs/base_zimage{,_turbo}.yml` | Inference configs |
| `tests/z_image/` | Transformer + module-parity tests |

## Performance (v6e-8, Z-Image-Turbo, 1024², 9 steps)

| | 8 images (`per_device_batch_size: 1.0`) | 1 image |
|---|---|---|
| generation | 2.29s — **0.28s/image** | **0.59s** |
| compile | ~110s | ~44s |
| load (warm Orbax cache) | ~12–20s | ~15s |

Default mesh is `data=8`. Measured alternatives at batch 8: `data=4/ctx=2` 0.325s/img, `ctx=8` 0.519s/img, `fsdp=8` 1.179s/img. 

## Notable fixes made along the way

- **`NNXFlaxQwen3Attention` was numerically broken** — no causal mask, and it read `attention_mask` as additive when callers pass 0/1. Cosine 0.242 vs HF Transformers; now 0.999996. Its existing test never checked numerics.
- **Qwen3 weights were stored fp32 under a bf16 config.** Flax stores params in `param_dtype` (float32 default) regardless of compute dtype, so the 4B encoder occupied 16.1 GB/chip instead of 8.0. Norm scales stay fp32 deliberately, now declared in the module.
- `get_git_commit_hash` / GCS upload / `delete_file` / `save_images` moved to `max_utils`, deduping ~90 lines across the `generate_wan`, `generate_ltx2` and `generate_zimage` drivers.

## Verification

- Qwen3 encoder vs HF Transformers on the real 4B checkpoint: **cosine 0.999996**.
- Orbax round trip on tiny models: param values, dtypes, target shardings, and a jitted batched generation.
- Full cold and warm runs on v6e-8; output is visually correct and matches the pre-refactor reference.
- `pyink --check` clean across 319 files; ruff clean on all touched files.

image

- We could extend max batchsize > 256, by offloading the Qwen Encoder.

### DIFF
--- a/src/maxdiffusion/models/z_image/transformer_z_image.py
+++ b/src/maxdiffusion/models/z_image/transformer_z_image.py
@@ -1,0 +1,962 @@
+"""
+Copyright 2026 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+"""JAX/NNX implementation of the Z-Image transformer."""
+
+from typing import Optional, Sequence
+
+import math
+
+import flax
+from flax import nnx
+import jax
+import jax.numpy as jnp
+import numpy as np
+
+from maxdiffusion.common_types import (
+    BATCH,
+    BlockSizes,
+    D_KV,
+    SELF_ATTN_HEAD,
+    SELF_ATTN_KV_LENGTH,
+    SELF_ATTN_Q_LENGTH,
+)
+from maxdiffusion.configuration_utils import ConfigMixin, register_to_config
+from maxdiffusion.models.attention_flax import NNXAttentionOp
+from .logical_sharding_z_image import get_sharding_specs, ZImageDiTShardingSpecs
+
+ADALN_EMBED_DIM = 256
+SEQ_MULTI_OF = 32
+
+
+def _linear(
+    rngs,
+    in_features,
+    out_features,
+    *,
+    use_bias=True,
+    dtype=jnp.float32,
+    weights_dtype=jnp.float32,
+    kernel_axes=("embed", "heads"),
+    bias_axes=("heads",),
+):
+  """Create a logically sharded NNX dense layer for the distributed denoiser."""
+  return nnx.Linear(
+      in_features,
+      out_features,
+      use_bias=use_bias,
+      rngs=rngs,
+      dtype=dtype,
+      param_dtype=weights_dtype,
+      kernel_init=nnx.with_partitioning(nnx.initializers.lecun_normal(), kernel_axes),
+      bias_init=nnx.with_partitioning(nnx.initializers.zeros_init(), bias_axes),
+  )
+
+
+@flax.struct.dataclass
+class ZImageTransformer2DModelOutput:
+  sample: list[jax.Array]
+
+
+class ZImageTimestepEmbedder(nnx.Module):
+  """Sinusoidal timestep embedding followed by the upstream two-layer MLP."""
+
+  def __init__(
+      self,
+      rngs: nnx.Rngs,
+      out_size: int,
+      mid_size: Optional[int] = None,
+      frequency_embedding_size: int = 256,
+      dtype: jnp.dtype = jnp.float32,
+      weights_dtype: jnp.dtype = jnp.float32,
+      sharding_specs: Optional[ZImageDiTShardingSpecs] = None,
+  ):
+    if sharding_specs is None:
+      sharding_specs = get_sharding_specs("default", "z_image_dit")
+    self.sharding_specs = sharding_specs
+    self.frequency_embedding_size = frequency_embedding_size
+    mid_size = out_size if mid_size is None else mid_size
+    self.mlp_in = _linear(
+        rngs,
+        frequency_embedding_size,
+        mid_size,
+        dtype=dtype,
+        weights_dtype=weights_dtype,
+        kernel_axes=self.sharding_specs.emb_linear_1_kernel,
+        bias_axes=self.sharding_specs.emb_linear_1_bias,
+    )
+    self.mlp_out = _linear(
+        rngs,
+        mid_size,
+        out_size,
+        dtype=dtype,
+        weights_dtype=weights_dtype,
+        kernel_axes=self.sharding_specs.emb_linear_2_kernel,
+        bias_axes=self.sharding_specs.emb_linear_2_bias,
+    )
+
+  @staticmethod
+  def timestep_embedding(timestep: jax.Array, dim: int, max_period: float = 10000.0) -> jax.Array:
+    half = dim // 2
+    freqs = jnp.exp(-math.log(max_period) * jnp.arange(half, dtype=jnp.float32) / half)
+    args = timestep[:, None].astype(jnp.float32) * freqs[None]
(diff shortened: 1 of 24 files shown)


src/maxdiffusion/pipelines/z_image/z_image_pipeline.py
```diff
@@ -0,0 +1,245 @@
...
+    # change the real tokens' states -- they are sliced back out below, and
+    # Z-Image consumes only those unmasked states.
+    inputs = self.tokenizer(
+        chats, padding="max_length", max_length=max_sequence_length, truncation=True, return_tensors="np"
+    )
+    graphdef, state, rest = self._text_encoder_split
+
+    if self.offload_encoders:
+      # Restore parameters from CPU back to original device shardings
+      state = jax.tree_util.tree_map(lambda x, sharding: jax.device_put(x, sharding), state, self._text_encoder_shardings)
```
Perseus14: Can we use `device_put_replicated`  (e.g., maxdiffusion.max_utils.device_put_replicated) instead of naive `device_put` tree map to ensure performant multi-host model swapping?

src/maxdiffusion/models/z_image/transformer_z_image.py
```diff
@@ -0,0 +1,805 @@
...
+    )
+
+  @staticmethod
+  def _apply_rope(x: jax.Array, freqs_cis: jax.Array, out_dtype: jnp.dtype) -> jax.Array:
+    """Apply complex RoPE; x is B,S,H,D and frequencies B,S,D/2."""
+    x_float = x.astype(jnp.float32).reshape(*x.shape[:-1], -1, 2)
+    real, imag = x_float[..., 0], x_float[..., 1]
+    cos, sin = jnp.real(freqs_cis)[:, :, None], jnp.imag(freqs_cis)[:, :, None]
+    out = jnp.stack((real * cos - imag * sin, real * sin + imag * cos), axis=-1)
+    return out.reshape(x.shape).astype(out_dtype)
```
Perseus14: Similar to how we do in LTX2 and WAN, we can precompute real-valued cos and sin tensors directly and pass them as real arrays instead of relying on JAX to unroll complex tensors during execution.

src/maxdiffusion/models/z_image/transformer_z_image.py
```diff
@@ -0,0 +1,805 @@
...
+  ):
+    if len(axes_dims) != len(axes_lens):
+      raise ValueError("axes_dims and axes_lens must have the same length.")
+    self.axes_dims = tuple(axes_dims)
+    self.axes_lens = tuple(axes_lens)
+    freqs = []
+    for dim, length in zip(self.axes_dims, self.axes_lens):
+      angular = 1.0 / (theta ** (np.arange(0, dim, 2, dtype=np.float32) / dim))
+      phase = np.arange(length, dtype=np.float32)[:, None] * angular[None]
+      freqs.append(np.exp(1j * phase).astype(np.complex64))
```
Perseus14: Instead of storing in `np.complex64` and then splitting in `_apply_rope`, we should refactor to use real `cos` and `sin` to be performant

src/maxdiffusion/pipelines/z_image/z_image_pipeline.py
```diff
@@ -0,0 +1,245 @@
...
+    if self.offload_encoders:
+      from maxdiffusion import max_logging
+
+      s_offload = time.perf_counter()
+      cpus = jax.devices("cpu")
+      offloaded_state = jax.tree_util.tree_map(lambda x: jax.device_put(x, device=cpus[0]), state)
+      self._text_encoder_split = (graphdef, offloaded_state, rest)
+      max_logging.log(f"Offloaded Qwen3 text encoder parameters to CPU in {(time.perf_counter() - s_offload):.4f}s.")
+
+    return [jnp.asarray(embeddings[index, : lengths[index]], dtype=self.dtype) for index in range(len(lengths))]
```
Perseus14: As we are slicing the embeddings to `lengths[index]`, the shape of these arrays will change depending on exactly how many tokens the user's prompt has. Since these arrays get passed directly into our `@jax.jit` compiled denoise_step, I am worried whether it will force a full (and very slow) XLA recompilation of the denoiser each time. 

To keep our serving loop blazing fast, I think we should avoid slicing here and instead pass the fixed-length padded tensors (e.g. [B, 512, D]) into the denoiser, along with an explicit boolean attention mask.

src/maxdiffusion/models/z_image/transformer_z_image.py
```diff
@@ -0,0 +1,805 @@
...
+      caption_positions.append(cap_position)
+      caption_masks.append(cap_mask)
+      image_features.append(patch_padded)
+      image_positions.append(image_position)
+      image_masks.append(image_mask)
+      sizes.append(size)
+
+    # 2. Find maximum lengths statically (Python values)
+    image_lengths = [item.shape[0] for item in image_features]
+    caption_lengths = [item.shape[0] for item in caption_features]
```
Perseus14: To prevent above mentioned jax recompilation, we should tweak this to accept an explicit `attention_mask` from the pipeline. We can use that mask to know exactly where the valid tokens end

Perseus14 (commented):
Thanks for the PR. This looks great! 

Added a few comments, PTAL.

Could you also run in on 7x and update the PR with the generation time and other details?

Additionally could we also implement the new optimizations from https://github.com/AI-Hypercomputer/maxdiffusion/pull/438 and https://github.com/AI-Hypercomputer/maxdiffusion/pull/443?

src/maxdiffusion/pipelines/z_image/z_image_pipeline.py
csgoogle: We are bucketing them in 32, so we only compute once, during compilation

src/maxdiffusion/pipelines/z_image/z_image_pipeline.py
csgoogle: thanks, done

src/maxdiffusion/models/z_image/transformer_z_image.py
csgoogle: thanks done

src/maxdiffusion/models/z_image/transformer_z_image.py
csgoogle: done

src/maxdiffusion/pipelines/z_image/z_image_pipeline.py
csgoogle: I checked again and fixed the text encoder output length so it defaults to 512 every time.

I kept the remaining padding logic since images within a batch can still be generated at different resolutions, though it won't be triggered in standard single-resolution runs.

src/maxdiffusion/models/z_image/transformer_z_image.py
csgoogle: Addressed above

Perseus14:
LGTM! Tested both zimage and zimage-turbo and I am able to generate images.

Thanks for the PR. Please fix the linting issues and I will approve the PR

Perseus14 (approved):
LGTM!

PR #450, opened 2026-07-23
Fixing nightly regression test

prishajain1:
Fix Wan 2.1 training nightly regression test failures by wrapping FlaxCLIPTextModel imports in max_utils.py with try...except ImportError to handle container environments with updated transformers releases.

### DIFF
--- a/src/maxdiffusion/max_utils.py
+++ b/src/maxdiffusion/max_utils.py
@@ -55,7 +55,11 @@ from flax.linen import partitioning as nn_partitioning
 from flax.training import train_state
 from jax.experimental import mesh_utils
 
-from transformers import FlaxCLIPTextModel, FlaxCLIPTextPreTrainedModel
+try:
+  from transformers import FlaxCLIPTextModel, FlaxCLIPTextPreTrainedModel
+except ImportError:
+  FlaxCLIPTextModel = None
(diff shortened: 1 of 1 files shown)


src/maxdiffusion/max_utils.py
```diff
@@ -55,7 +55,11 @@
 from flax.training import train_state
 from jax.experimental import mesh_utils
 
-from transformers import FlaxCLIPTextModel, FlaxCLIPTextPreTrainedModel
+try:
+  from transformers import FlaxCLIPTextModel, FlaxCLIPTextPreTrainedModel
+except ImportError:
+  FlaxCLIPTextModel = None
+  FlaxCLIPTextPreTrainedModel = None
```
Perseus14, line 62: Shouldn't we import these from somewhere else otherwise it might lead to broken code?

prishajain1: The reason they are wrapped in a try...except ImportError block in max_utils.py is due to compatibility issues in nightly container environments (maxdiffusion_jax_stable_stack_nightly), where the transformers library fails to load Flax modules due to JAX version mismatches.

This only prevents startup crashes for models like Wan 2.1, which import max_utils.py but never load those CLIP-specific pipelines.

Perseus14 (approved):

PR #455, opened 2026-07-31
Fix multi-host profiling and resolve profiler tracer collisions

richaguptaa17:
- Enable ML Diagnostics profiler on secondary TPU worker processes by removing the jax.process_index() == 0 lock from _ml_diagnostics_profiler_enabled().
- Refactor Profiler class in max_utils.py with active state tracking (_active). When both enable_ml_diagnostics and enable_profiler are True, prioritize ML Diagnostics and skip the standalone JAX profiler with a log warning to avoid RuntimeError tracer collisions.

### DIFF
--- a/src/maxdiffusion/max_utils.py
+++ b/src/maxdiffusion/max_utils.py
@@ -85,7 +85,7 @@ def _jax_profiler_enabled(config):
 
 
 def _ml_diagnostics_profiler_enabled(config):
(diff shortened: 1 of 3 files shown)


mbohlool (approved):

entrpn (approved):

src/maxdiffusion/max_utils.py
sagarchapara, line 164: In a multi-host setup, won't non-master hosts upload their trace files to the same GCS prefix and overwrite each other?

PR #464, opened 2026-08-19
Improvement in WAN VAE Performance

Toshi-31:
### Improvement in WAN VAE Performance

Changes made:
1. Using jnp.repeat in WAN Upsample instead of resize
2. VAE Spatial and Temporal Parallelism (vae_spatial=8 & vae_decode_chunk=5)
3. Host Formatting Changes: Pushes Classifier-Free Guidance (CFG) tensor duplication inside the JIT-compiled device graph to completely eliminate Python host overhead.

Performance Impact:
Tested on v6e-8 for 720p, 81-frame video. These changes reduced step timings from:

Before:
- VAE Decode: 3.2s
  - TPU Compute: 2.8s
  - Host Formatting: 0.4s

After:
- VAE Decode: 0.8s
  - TPU Compute: 0.8s
  - Host Formatting: 0.0s

Leading to a 75% improvement in latency.

For v7x-8, the observations are below. The improvement is not as significant as that in v6e-8.
Before:
VAE Decode:              3.6s
    - TPU Compute:         2.9s
    - Host Formatting:     0.7s
After:
VAE Decode:              2.9s
    - TPU Compute:         2.9s
    - Host Formatting:     0.0s

### DIFF
--- a/src/maxdiffusion/models/wan/autoencoder_kl_wan.py
+++ b/src/maxdiffusion/models/wan/autoencoder_kl_wan.py
@@ -23,6 +23,7 @@ from jax.sharding import NamedSharding, PartitionSpec as P
 from jax import tree_util
 from flax import nnx
 from ...configuration_utils import ConfigMixin
+from ... import max_logging
 from ..modeling_flax_utils import FlaxModelMixin, get_activation
 from ... import common_types
 from ..vae_flax import (
@@ -99,9 +100,9 @@ class WanCausalConv3d(nnx.Module):
     self.mesh = mesh
 
     # Weight sharding (Kernel is sharded along output channels)
-    num_fsdp_devices = mesh.shape["vae_spatial"]
+    num_fsdp_devices = mesh.shape["vae_spatial"] if mesh is not None and "vae_spatial" in mesh.shape else 1
     kernel_sharding = (None, None, None, None, None)
-    if out_channels % num_fsdp_devices == 0:
+    if num_fsdp_devices > 1 and out_channels % num_fsdp_devices == 0:
       kernel_sharding = (None, None, None, None, "vae_spatial")
 
     self.conv = nnx.Conv(
@@ -119,8 +120,10 @@ class WanCausalConv3d(nnx.Module):
     )
 
   def __call__(self, x: jax.Array, cache_x: Optional[jax.Array] = None, idx=-1) -> jax.Array:
-    spatial_sharding = NamedSharding(self.mesh, P("redundant", None, None, "vae_spatial", None))
-    x = jax.lax.with_sharding_constraint(x, spatial_sharding)
+    if self.mesh is not None and "vae_spatial" in self.mesh.shape:
+      spatial_sharding = NamedSharding(self.mesh, P("redundant", None, None, "vae_spatial", None))
+      if spatial_sharding is not None:
+        x = jax.lax.with_sharding_constraint(x, spatial_sharding)
 
     current_padding = list(self._causal_padding)
     padding_needed = self._depth_padding_before
@@ -198,8 +201,16 @@ class WanUpsample(nnx.Module):
     n, h, w, c = in_shape
     target_h = int(h * self.scale_factor[0])
     target_w = int(w * self.scale_factor[1])
-    out = jax.image.resize(x.astype(jnp.float32), (n, target_h, target_w, c), method=self.method)
-    return out.astype(input_dtype)
+    if self.method == "nearest" and self.scale_factor[0] == int(self.scale_factor[0]) and self.scale_factor[1] == int(self.scale_factor[1]):
+      scale_h = int(self.scale_factor[0])
+      scale_w = int(self.scale_factor[1])
+      out = jnp.repeat(jnp.repeat(x, scale_h, axis=1), scale_w, axis=2)
+    else:
+      if self.method == "nearest":
+        max_logging.log(f"Warning: WanUpsample2D nearest method requested but scale_factor {self.scale_factor} is not integer. Falling back to jax.image.resize.")
+      out = jax.image.resize(x.astype(jnp.float32), (n, target_h, target_w, c), method=self.method)
+      out = out.astype(input_dtype)
+    return out
 
 
 class Identity(nnx.Module):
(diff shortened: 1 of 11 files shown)


src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -67,6 +67,28 @@ def __eq__(self, other):
...
+      if len(spec) != x.ndim:
+        return x
+      for axis_idx, axis_names in enumerate(spec):
+        if axis_names is not None:
+          if not isinstance(axis_names, tuple):
+            axis_names = (axis_names,)
+          mesh_axis_size = 1
+          for axis_name in axis_names:
+            if axis_name in mesh.shape:
+              mesh_axis_size *= mesh.shape[axis_name]
```
csgoogle: we can log warnings when they mismatch

src/maxdiffusion/pipelines/wan/wan_pipeline.py
```diff
@@ -907,8 +939,10 @@ def _decode_latents_to_video(self, latents: jax.Array, trace: Optional[dict] = N
     if trace is not None:
       trace["vae_decode_tpu"] = time.perf_counter() - t_vae_tpu_start
 
-    video = jax.experimental.multihost_utils.process_allgather(video, tiled=True)
```
csgoogle, line 910: Is this change expected?

Toshi-31: Yes, it changes the python-side gather operation and replaces it with hardware-level synchronization.

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
Toshi-31: Sure, will do

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -84,6 +84,8 @@ def _with_sharding_constraint(x, sharding):
             if axis_name in mesh.shape:
               mesh_axis_size *= mesh.shape[axis_name]
           if x.shape[axis_idx] % mesh_axis_size != 0:
+            import logging
```
Perseus14: Let's use our custom `max_logging` here.

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -221,11 +223,14 @@ def __call__(self, x: jax.Array) -> jax.Array:
...
     target_h = int(h * self.scale_factor[0])
     target_w = int(w * self.scale_factor[1])
-    if self.method == "nearest":
+    if self.method == "nearest" and self.scale_factor[0] == int(self.scale_factor[0]) and self.scale_factor[1] == int(self.scale_factor[1]):
       scale_h = int(self.scale_factor[0])
       scale_w = int(self.scale_factor[1])
       out = jnp.repeat(jnp.repeat(x, scale_h, axis=1), scale_w, axis=2)
     else:
+      if self.method == "nearest":
+        import logging
```
Perseus14: Same as above

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -1131,7 +1167,6 @@ def __init__(
     )
     self.mesh = mesh
 
-  @nnx.jit
```
Perseus14, line 1134: We might need to keep this as WAN VACE directly uses `_encode`

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -1239,7 +1279,6 @@ def encode(
       return (posterior,)
     return FlaxAutoencoderKLOutput(latent_dist=posterior)
 
-  @nnx.jit
```
Perseus14, line 1242: Same as above

Perseus14 (commented):
Few minor changes requested.

Please squash the commits.

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
Toshi-31: I've updated the VACE pipeline to use a jitted wrapper (vae_encode_pass) instead of calling self.vae.encode directly.

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
Toshi-31: Sure, will do

Perseus14 (dismissed):
LGTM!

@ninatu Could you also PTAL?

ninatu (dismissed):
LGTM

Perseus14 (dismissed):

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
```diff
@@ -67,6 +68,29 @@ def __eq__(self, other):
 tree_util.register_pytree_node(RepSentinel, lambda x: ((), None), lambda _, __: RepSentinel())
 
 
+def _with_sharding_constraint(x, sharding):
+  if sharding is not None and hasattr(x, "shape"):
```
mbohlool: why do you need these two if statements? does jax.lax.with_sharding constraint silency gnore null sharding or x with no shape or sharding with no mesh an spec?

Toshi-31: Yes, jax.lax.with_sharding_constraint strictly requires a valid PartitionSpec/Sharding object and a valid JAX array. It does not silently ignore None or non-array types; it will raise a TypeError if either is invalid.

Since we apply this constraint to multiple intermediate tensors across the VAE blocks, we abstracted it into this helper function. This centralizes the safety checks (ensuring it doesn't crash when spatial sharding is disabled or when encountering optional None tensors) and avoids scattering identical if boilerplate everywhere in the forward pass.

mbohlool: Most of the places I see you use this function is to replace  jax.lax.with_sharding_constraint, right? What I am concern about is silence failure. in current code,  jax.lax.with_sharding_constraint expect those attributes to be there and if not, it will fail and we will see the failure (and as we don't see any failure now, that means current code is passing the right attributes). When we replace it with this function, if in future a bug resulted in missing those attributes, it will silently ignored here which is not good. I suggest you remove these checks here assuming they are there, and if in any call site, you want to ignore it, add check there with right comment. most of the replacements of  jax.lax.with_sharding_constraint with this function should not ignore it (I think all of them should not).

mbohlool (changes requested):
Please address the comment above.

src/maxdiffusion/models/wan/autoencoder_kl_wan.py
Toshi-31: Done!

Perseus14 (approved):

PR #470, opened 2026-09-01
Wan 2.2 training

Toshi-31:
**Title:**
Add Wan 2.2 Training Pipeline with Joint Timestep Routing

**Description:**
This PR introduces full training support for the Wan 2.2 models into MaxDiffusion.

**1. Joint Timestep Routing**
This implementation utilizes a unified, joint-pipeline training strategy. During each training step, the pipeline dynamically samples a timestep and routes the forward pass to either the high-noise or low-noise transformer based on the configured boundary ratio:

is_high_noise = jax.random.uniform(cond_rng) > config.boundary_ratio
This ensures a seamless and efficient training execution graph without needing separate high/low training loops.

**2. Pipeline Inheritance & Code Reuse**
The new wan_trainer_2_2.py relies directly on the WanPipeline2_2 class to bootstrap the models and load checkpoints. Because WanPipeline2_2 cleanly inherits from the base WanPipeline, the training loop seamlessly reuses all existing weight conversion and fast-loading logic that was previously established for inference.

The loss graphs were plotted for around 260 steps, and show a clear downward trend.

For graphs and other artefacts: https://docs.google.com/document/d/1svzC8cVZxb2XxyypeFcoJig13_1ptu6wYIe5QwnC_Lo/edit?usp=sharing

**Step time**: Around 39.7 seconds per device. 

Through systematic roofline analysis and iterative kernel, parallelism, rematerialization, and collective optimizations (from V2 through V7), steady-state step time is reduced from ~39.7 s → 13.57 s (2.92× end-to-end speedup, -65.8% step time reduction), increasing per-device throughput from 131.6 → 384.12 TFLOP/s/device and driving Model FLOP Utilization (MFU) from ~7.5% → 40.18%.

Note on Code Changes: This PR introduces the training recipe, parallelism mesh, and XLA/Pallas config updates. All underlying kernel and model code changes (custom Pallas splash attention backward kernel with aliased dQ reduction, remat checkpoint tagging, and block-size wiring) will be raised in a companion PR in AI-Hypercomputer/maxdiffusion.

Optimization Summary & Speedup
| Optimization | Description | Impact / Speedup |
|--------|--------|--------|
| 1. Selective Activation Remat & Host Overhead Removal (V2–V3) | Switched from full remat (`remat_policy=FULL`) to selective checkpointing (`MATMUL_WITHOUT_BATCH`), keeping linear and attention activations in HBM (~7.7 GB across 40 layers) while recomputing only elementwise ops (RMSNorm, RoPE). Removed hot-path 14B parameter PyTree scans and host-device step counter synchronizations. | Eliminated 640 redundant forward GEMMs and attention calls per step. Reduced step time from 39.7 s → 28.50 s (-28.2%, +51.35 TFLOP/s/device). |
| 2. Pure 1D FSDP Parallelism & SparseCore Offload (V4) | Reconfigured hybrid parallelism (`dp=2, fsdp=8`) to pure 1D FSDP (`dp=1, fsdp=16, cp=4, tp=1`), offloading 3D All-Gather and ND Reduce-Scatter torus collective transfers to the 32 SparseCore coprocessors while bounding collective concurrency. | 100% eliminated the synchronous 27B parameter DP All-Reduce phase (saving 6.05 s of stalls). Halved model weights and Adam states in HBM (saving ~8.5 GB), reducing step time to 27.63 s. |
| 3. Overlapped Ring Attention & Sequence Sharding (V4.2) | Replaced global sequence all-gathers (`attention=tokamax`) with Overlapped Ring Attention (`attention=tokamax_ring`), keeping Key and Value states sharded locally at 18,900 tokens instead of replicating the full 75,600 sequence, and overlapped ring permutations behind compute. | Resolved compilation HBM OOM (required 112.7 GB vs 94.7 GB available), eliminated 160 synchronous KV all-gathers (~3.5 s) and the duplicate `splash_mha_dq` pass. Dropped step time to 22.38 s (232.85 TFLOP/s). |
| 4. High-Intensity FlashAttention Tile Scaling & Layer Scheduling (V4.1–V4.4) | Scaled FlashAttention compute tiles from 1024 to 2048 (`block_q=2048, block_kv=2048, block_kv_compute=1024`) with native VPU base-2 exponentiation (`use_base2_exp=True`) to saturate TPU v7x 256×256 systolic arrays, and re-enabled Latency-Hiding Layer Scheduling (LHS) with a strict 105% memory limit. | Halved inner-loop stepping overhead, doubled MXU burst lengths, and overlapped inter-layer FSDP All-Gathers/Reduce-Scatters with compute. Decreased step time to 21.82 s (238.8 TFLOP/s). |
| 5. Global All-Reduce Elimination & Unsegmented Attention Backward (V5) | Disabled Optax global L2 gradient norm clipping (`opt_enable_grad_global_norm_clipping=False`) to avoid blocking cross-host barrier synchronization on every step. Disabled padding token masking (`mask_padding_tokens=False`) to dispatch to `splash_mha_dkv_no_residuals`, eliminating vector unit boundary mask evaluations. | Dropped collective All-Reduce overhead from 6.0 s down to 31 ms. Reduced backward attention latency from 7.40 s to 6.59 s, driving step time down to 19.37 s (269.1 TFLOP/s, 14.95% MFU). |
| 6. Duplicate Forward Attention Remat Elimination & Chunked Ulysses (V6) | Identified and eliminated duplicate forward attention recompute in the backward pass by tagging `out` and `lse` with `ad_checkpoint.checkpoint_name("attn_output")` in `ring_attention_kernel.py` and saving it in `remat_policy=CUSTOM`. Chunked Ulysses context parallelism into 2 chunks of 5 heads (`ulysses_attention_chunks=2`) with async All-to-All pipelining. | Eliminated 80 duplicate `splash_mha_fwd` calls per step (~3.96 s savings), reduced peak HBM by 6.96 GB, and lowered step time from 19.37 s → 15.02 s (347.18 TFLOP/s, 36.31% MFU). |
| 7. Custom Splash Attention Backward Kernel: Aliased dQ & VPU Sub-Tiling (V7) | Replaced the 37-deep unaliased FP32 `dQ` scratch tensor (7.18 GB/layer) with a 3-slot ring buffer (`dq_reduction_steps=3`, `input_output_aliases={6: 0}`) with async DMA prefetching and in-VMEM accumulation, added inner VPU sub-tiling (`bkv_compute_in=256`), and fused attention output reciprocal normalization (`fuse_reciprocal=True`). | Reduced `dQ` scratch HBM traffic by 12.3× (7.18 GB → 0.58 GB/layer) and cut backward attention kernel time by 33.7% (4.24 s → 2.81 s/step). Reduced step time from 15.02 s → 13.57 s (-1.45 s, 384.12 TFLOP/s, 40.18% MFU). |

## Key Progression Milestones
Baseline (V1): 39.70 s / step | 131.6 TFLOP/s | ~7.5% MFU
V3 (Selective Remat): 28.50 s / step | 182.95 TFLOP/s | 10.2% MFU
V5 (Unsegmented Attention + No All-Reduce): 19.37 s / step | 269.1 TFLOP/s | 14.95% MFU
V6 (Remat Fix + Chunked Ulysses): 15.02 s / step | 347.18 TFLOP/s | 36.31% MFU
V7 (Custom Splash Bwd Kernel): 13.57 s / step | 384.12 TFLOP/s | 40.18% MFU (2.92× total speedup)


src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,558 @@
...
+          mesh,
+          config.global_batch_size_to_load,
+          pipeline=pipeline,
+          is_training=is_training,
+      )
+
+    if config.dataset_type != "tfrecord" and not config.cache_latents_text_encoder_outputs:
+      raise ValueError(
+          "Wan 2.2 training only supports config.dataset_type set to tfrecords and config.cache_latents_text_encoder_outputs set to True"
+      )
```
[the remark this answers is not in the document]
Toshi-31: Fixed!

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,558 @@
...
+      max_logging.log(f"Step {step}, Final Average Eval loss: {final_eval_loss:.4f}")
+      if writer:
+        writer.add_scalar("learning/eval_loss", final_eval_loss, step)
+
+
+def train_step_2_2(state_low, state_high, data, rng, scheduler_state, scheduler, config):
+  _, new_rng, timestep_rng, dropout_rng, cond_rng, noise_rng = jax.random.split(rng, num=6)
+  
+  for k, v in data.items():
+    data[k] = v[: config.global_batch_size_to_train_on, :]
```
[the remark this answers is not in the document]
Toshi-31: Done

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,558 @@
...
+    _, new_rng = jax.random.split(rng, num=2)
+    
+    is_high = timesteps[0] >= boundary
+    
+    def true_fn(_):
+      return loss_fn_high(state_high.params, latents, encoder_hidden_states, timesteps, new_rng)
+    def false_fn(_):
+      return loss_fn_low(state_low.params, latents, encoder_hidden_states, timesteps, new_rng)
+      
+    loss = jax.lax.cond(is_high, true_fn, false_fn, operand=None)
```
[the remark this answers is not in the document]
Toshi-31: Fixed

src/maxdiffusion/pyconfig.py
```diff
@@ -284,11 +284,11 @@ def user_init(raw_keys):
...
     raw_keys["tokenizer_model_name_or_path"] = raw_keys["pretrained_model_name_or_path"]
     if "gs://" in raw_keys["pretrained_model_name_or_path"]:
-      raw_keys["pretrained_model_name_or_path"] = max_utils.download_blobs(raw_keys["pretrained_model_name_or_path"], "/tmp")
+      raw_keys["pretrained_model_name_or_path"] = max_utils.download_blobs(raw_keys["pretrained_model_name_or_path"], "/dev/shm")
     if "gs://" in raw_keys["unet_checkpoint"]:
-      raw_keys["unet_checkpoint"] = max_utils.download_blobs(raw_keys["unet_checkpoint"], "/tmp")
+      raw_keys["unet_checkpoint"] = max_utils.download_blobs(raw_keys["unet_checkpoint"], "/dev/shm")
     if "gs://" in raw_keys["tokenizer_model_name_or_path"]:
-      raw_keys["tokenizer_model_name_or_path"] = max_utils.download_blobs(raw_keys["tokenizer_model_name_or_path"], "/tmp")
+      raw_keys["tokenizer_model_name_or_path"] = max_utils.download_blobs(raw_keys["tokenizer_model_name_or_path"], "/dev/shm")
```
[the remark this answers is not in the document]
Toshi-31: I am running this on TPU VMs, where /dev/shm utilizes half of the system RAM (providing hundreds of GBs of space). The OS root disk (/tmp) on TPU VMs is extremely small (~100GB). When downloading the 50GB+ weights to /tmp, it entirely exhausts the OS disk and causes the Kubernetes pod to crash with a SIGTERM. Pointing to /dev/shm is an intentional fix for TPU VMs.

src/maxdiffusion/models/wan/wan_utils.py
```diff
@@ -498,20 +498,32 @@ def convert_chunk(ckpt_shard_path, chunk_keys):
...
-  )
-  with concurrent.futures.ThreadPoolExecutor(max_workers=32) as executor:
-    futures = [executor.submit(convert_chunk, path, keys) for path, keys in tasks]
-    for future in concurrent.futures.as_completed(futures):
-      future.result()  # re-raise conversion errors
+      
+    with concurrent.futures.ThreadPoolExecutor(max_workers=32) as executor:
+      futures = [executor.submit(convert_chunk, path, keys) for path, keys in tasks]
+      for future in concurrent.futures.as_completed(futures):
+        future.result()  # re-raise conversion errors
```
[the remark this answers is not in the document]
Toshi-31: Done

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,558 @@
...
+
+    pipeline = self.training_loop_2_2(
+        pipeline, optimizer_low, optimizer_high, learning_rate_scheduler, train_data_iterator, restore_args
+    )
+
+    if self.config.enable_ssim:
+      posttrained_video_path = self.generate_sample(self.config, pipeline, filename_prefix="post-training-")
+      print_ssim(pretrained_video_path, posttrained_video_path)
+
+  def training_loop_2_2(self, pipeline, optimizer_low, optimizer_high, learning_rate_scheduler, train_data_iterator, restore_args: dict = {}):
```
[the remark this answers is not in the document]
Toshi-31: Fixed

prishajain1:
Can you please add the current timestep in the PR description

src/maxdiffusion/configs/base_wan_27b.yml
```diff
@@ -225,9 +226,9 @@ vae_logical_axis_rules: [
                           ['heads', null],
                           ['norm', null],
                           ['conv_batch', 'redundant'],
-                          ['out_channels', 'vae_spatial'],
-                          ['conv_out', 'vae_spatial'],
-                          ['conv_in', 'vae_spatial'],
+                          ['out_channels', null],
+                          ['conv_out', null],
+                          ['conv_in', null],
```
csgoogle: we could add a config file for training? this is base config and would impact inference too right?

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,554 @@
...
+          "learning/max_grad_norm_low": max_grad_norm_low,
+          "learning/max_grad_norm_high": max_grad_norm_high,
+          "learning/max_abs_grad_low": max_abs_grad_low,
+          "learning/max_abs_grad_high": max_abs_grad_high,
+      },
+      "scalars": {},
+  }
+
+  new_state_low = state_low.apply_gradients(grads=grads_low)
+  new_state_high = state_high.apply_gradients(grads=grads_high)
```
prishajain1: We can apply gradient conditionally to the active transformer instead of calling it on both always

setup.sh
prishajain1, line 1: Any reason why changes in setup.sh, docker build, and maxdiffusion dep dockerfile have been included in this PR? If these were local debugging workarounds, lets remove them. If these are necessary, I suggest to include these in a separate PR. This might also have the risk of breaking nightly image build process. Let's scope this PR to wan 2.2 training alone

prishajain1:
Please add relevant unittests

src/maxdiffusion/pyconfig.py
prishajain1: Instead of hardcoding /dev/shm, you could add a config like checkpoint_save_location in the base yaml file with default value of /tmp, and use that variable here

src/maxdiffusion/models/wan/wan_utils.py
```diff
@@ -498,20 +498,32 @@ def convert_chunk(ckpt_shard_path, chunk_keys):
...
-  tasks = []
-  for model_file in model_files:
-    ckpt_shard_path = resolve_shard_path(model_file)
-    with safe_open(ckpt_shard_path, framework="pt") as f:
-      shard_keys = [k for k in f.keys() if "norm_added_q" not in k]
-    for i in range(0, len(shard_keys), chunk_size):
-      tasks.append((ckpt_shard_path, shard_keys[i : i + chunk_size]))
   max_logging.log(
-      f"Load and port {pretrained_model_name_or_path} {subfolder}: {len(model_files)} shards, {len(tasks)} chunks"
+      f"Load and port {pretrained_model_name_or_path} {subfolder}: {len(model_files)} shards sequentially to save disk"
```
prishajain1: Currently the for loop finishes for all shards and then the deletion for that shard is happening, this could be a misplaced indent?

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,554 @@
...
+              state_low, state_high, example_batch, rng, scheduler_state
+          )
+          train_metric["scalar"]["learning/loss"].block_until_ready()
+        last_step_completion = datetime.datetime.now()
+
+        train_utils.record_scalar_metrics(
+            train_metric, last_step_completion - start_step_time, per_device_tflops, learning_rate_scheduler(step)
+        )
+        
+        loss_high_val = np.array(train_metric["scalar"]["learning/loss_high"])
```
prishajain1: do we need conversion to np array here as this causes d2h stalls, if this is for pure tensorboard plotting purpose can u figure out a workaround

src/maxdiffusion/configs/base_wan_27b.yml
Toshi-31: Done! Added training_wan_27b.yml.

Toshi-31:
> Can you please add the current timestep in the PR description

Done!

Toshi-31:
> Please add relevant unittests

Done!

src/maxdiffusion/trainers/wan_trainer_2_2.py
Toshi-31: Yeah, fixed

src/maxdiffusion/pyconfig.py
Toshi-31: Done

src/maxdiffusion/models/wan/wan_utils.py
Toshi-31: Yeah it was a misplaced indent. FIxed.

src/maxdiffusion/trainers/wan_trainer_2_2.py
Toshi-31: Yes it was meant for plotting purpose only. Removing it from here.

setup.sh
Toshi-31: Yes they were dubugging workarounds. Removing them.

src/maxdiffusion/configs/training_wan_27b.yml
Perseus14, line 1: Do we need a separate config file for training?

Toshi-31: Ref: https://github.com/AI-Hypercomputer/maxdiffusion/pull/470#discussion_r3903866963

The initial reason training_wan_27b.yml was added was to override the VAE convolution mesh axes (conv_in / conv_out) to null to prevent spatial indivisibility errors during training encode on small batch sizes.

Instead of creating a duplicate config file, I have now modified base_wan_27b.yml to keep replicate_vae: False as default (matching base_wan_14b.yml), and during training we pass replicate_vae=True (or vae_spatial=1) as a CLI parameter. When replicate_vae=True, the model automatically overrides the VAE spatial sharding without needing any separate YAML file.

training_wan_27b.yml has been removed.

src/maxdiffusion/configs/base_wan_27b.yml
```diff
@@ -409,6 +409,9 @@ num_inference_steps: 40
 fps: 16
 save_final_checkpoint: False
 
+# Location to download pretrained weights
+checkpoint_save_location: "/tmp"
```
entrpn, line 416: what is the purpose of this flag and how does it differ from pretrained_model_name_or_path?

Toshi-31: Earlier the check point save location was hardcoded as /tmp, which lead to No Space left on Device error for large models. I have set this as a configurable parameter. 
https://github.com/AI-Hypercomputer/maxdiffusion/pull/470/changes#diff-43ca5f8f5edd7a6ef8da8c66c1ae4e8c38b0322bb605afdfcdf51a7bc3ec1327L289

src/maxdiffusion/train_wan_2_2.py
prishajain1, line 1: why do we need a separate file? intended way of running wan2.2 training should be `python train_wan.py base_wan_27b.yml ....`
Please evaluate if this additional file is needed or it's logic can be included in train_wan.py (similar to how we have a single generate_wan.py for all wan inference)

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,721 @@
...
+  data = {k: v[: config.global_batch_size_to_train_on] for k, v in data.items()}
+
+  bsz = data["latents"].shape[0]
+  num_train_timesteps = scheduler.config.num_train_timesteps
+  boundary = int(config.boundary_ratio * num_train_timesteps)
+
+  is_high_noise = jax.random.uniform(cond_rng) > config.boundary_ratio
+
+  def compute_loss_high(high_params, s_high):
+    timesteps = jax.random.randint(timestep_rng, shape=(bsz,), minval=boundary, maxval=num_train_timesteps)
```
prishajain1, line 509: Do we not need to sample timesteps by applying time shift before scaling like other trainers in the repo do? using https://github.com/AI-Hypercomputer/maxdiffusion/blob/main/src/maxdiffusion/schedulers/scheduling_flow_match_flax.py#L247

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,721 @@
...
+
+    max_logging.log(f"Calculated TFLOPs per pass: {train_tflops:.4f}")
+    return train_tflops, total_attn_flops, seq_len
+
+  def get_train_step(self, pipeline, mesh, state_shardings, data_shardings):
+    return jax.jit(
+        functools.partial(train_step_2_2, scheduler=pipeline.scheduler, config=self.config),
+        in_shardings=(state_shardings["low_noise"], state_shardings["high_noise"], data_shardings, None, None),
+        out_shardings=(state_shardings["low_noise"], state_shardings["high_noise"], None, None, None),
+    )
```
prishajain1, line 163: donate_argnums=(0, 1)?

src/maxdiffusion/train_utils.py
```diff
@@ -130,7 +130,10 @@ def _tensorboard_writer_worker(writer, config):
     metrics, step = data
     if jax.process_index() == 0:
       for metric_name in metrics.get("scalar", []):
-        writer.add_scalar(metric_name, np.array(metrics["scalar"][metric_name]), step)
+        val = np.array(metrics["scalar"][metric_name])
+        if not np.isnan(val):
```
prishajain1, line 134: Could this filtering live in the wan 2.2 trainer instead? train_utils is shared, and dropping NaNs globally means a real divergence in any model just quietly disappears from Tensorboard.

src/maxdiffusion/checkpointing/wan_checkpointer_2_2.py
```diff
@@ -96,6 +126,7 @@ def config_to_json(model_or_config):
     max_logging.log(f"Saving checkpoint for step {train_step}")
     items = {
         "wan_config": ocp.args.JsonSave(config_to_json(pipeline.low_noise_transformer)),
+        "wan_config_high": ocp.args.JsonSave(config_to_json(pipeline.high_noise_transformer)),
```
prishajain1, line 129: Is this used anywhere? from_checkpoint still builds both transformers from wan_config, so this gets saved and restored but never read. I'd either wire it into create_model for the high-noise transformer, or leave it out until something needs it.

prishajain1 (changes requested):

src/maxdiffusion/trainers/wan_trainer_2_2.py
Toshi-31: Because of the conditional routing in lax.cond, only one expert is updated per step while the other is returned untouched. Donating a buffer that is returned unmodified causes XLA buffer invalidation errors or forces defensive memory copies. Since JAX does not support conditional buffer donation, I omitted donate_argnums to keep both model states stable in memory.

src/maxdiffusion/train_wan_2_2.py
Toshi-31: Sure, done!

src/maxdiffusion/trainers/wan_trainer_2_2.py
Toshi-31: Yes, applied time shift before scaling...

src/maxdiffusion/train_utils.py
Toshi-31: I've removed the global NaN filtering from train_utils.py so genuine divergences aren't suppressed. The inactive-expert and empty-bucket NaNs are now filtered locally inside wan_trainer_2_2.py before metrics are recorded.

src/maxdiffusion/checkpointing/wan_checkpointer_2_2.py
Toshi-31: Wired it into create_sharded_logical_transformer in wan_pipeline.py so that when subfolder == 'transformer' (high noise) , wan_config_high is read when available (falling back to wan_config), and for low noise, wan_config is read

src/maxdiffusion/trainers/wan_trainer_2_2.py
prishajain1: train step shifts t before expert split and applies shift afterwards, but inference routes on shifted timestamp? Can you pls check this and correct

Toshi-31: Yes I checked..previously I was partitioning unshifted continuous time u at boundary_ratio before applying the time shift, which shifted the boundary to ~0.988 rather than aligning with the inference threshold (boundary_ratio * 1000 = 875).

I've updated the training pipeline:

In train_step_2_2, we now compute the unshifted boundary using the inverse time shift: u_boundary = boundary_ratio / (shift - (shift - 1) * boundary_ratio).
The high-noise expert samples u in [u_boundary, 1.0] and the low-noise expert samples u in [0.0, u_boundary], so that after applying the time shift, their generated timesteps cover exactly [875, 1000] and [0, 875], matching inference.
The routing probability and optimizer schedule steps are scaled by (1 - u_boundary) and u_boundary.

prishajain1 (approved):

prishajain1 (dismissed):

Perseus14 (dismissed):

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,764 @@
...
+    eval_losses_by_timestep = {}
+    while True:
+      try:
+        eval_start_time = datetime.datetime.now()
+        eval_batch = load_next_batch(eval_data_iterator, None, self.config)
+        with mesh, nn_partitioning.axis_rules(self.config.logical_axis_rules):
+          metrics, eval_rng = p_eval_step(state_low, state_high, eval_batch, eval_rng, scheduler_state)
+          metrics["scalar"]["learning/eval_loss"].block_until_ready()
+
+        losses = jax.device_get(metrics["scalar"]["learning/eval_loss"])
```
mbohlool: 🔴 **Blocker**: Missing multi-host gather discards remote host evaluation metrics.

In `eval_2_2`, `eval_batch` is partitioned across the mesh's data axis (`NamedSharding(mesh, P(*self.config.data_sharding))`). On multi-host TPU topologies (e.g. v4-16, v5e-16+), each process computes losses on its local data slice only.

Because `multihost_utils.process_allgather` was omitted here, host 0's loop (`if jax.process_index() == 0:`) only records host 0's local slice and discards all evaluation samples from the other hosts. Additionally, calling `jax.device_get` directly on an array sharded across remote hosts can trigger JAX addressability exceptions.

Please add `multihost_utils.process_allgather(..., tiled=True)` before `jax.device_get`, matching [`BaseWanTrainer.eval`](file:///mnt/workspace/maxdiffusion/src/maxdiffusion/trainers/base_wan_trainer.py#L224-L227):

```suggestion
        from jax.experimental import multihost_utils

        losses = metrics["scalar"]["learning/eval_loss"]
        timesteps = metrics["scalar"]["learning/eval_timesteps"]
        gathered_losses = jax.device_get(multihost_utils.process_allgather(losses, tiled=True))
        gathered_timesteps = jax.device_get(multihost_utils.process_allgather(timesteps, tiled=True))

        if jax.process_index() == 0:
          for t, l in zip(gathered_timesteps.flatten(), gathered_losses.flatten()):
```

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,764 @@
...
+          self._profiler.start()
+        step_rng = jax.random.fold_in(base_rng, step_int)
+        start_step_time = datetime.datetime.now()
+        next_batch_future = executor.submit(load_next_batch, train_data_iterator, example_batch, self.config)
+
+        # Track active expert on host without device-to-host sync stall
+        _, _, _, _, cond_rng, _ = jax.random.split(step_rng, num=6)
+        shift = getattr(pipeline.scheduler.config, "shift", 1.0) or 1.0
+        u_boundary = compute_unshifted_boundary(self.config.boundary_ratio, shift)
+        is_high_noise = bool(jax.random.uniform(cond_rng) >= u_boundary)
```
mbohlool, line 394: 🔴 **Blocker**: Dual independent random evaluation risks host/device routing divergence.

Currently, `is_high_noise` is evaluated separately on CPU (here at L394) and inside the compiled TPU device graph (at L545):
- **Host** (L394): `is_high_noise = bool(jax.random.uniform(cond_rng) >= u_boundary)`
- **Device** (L545): `is_high_noise = jax.random.uniform(cond_rng) >= u_boundary`

Because CPU and TPU can have minor rounding or fast-math variations near $u_{\text{boundary}}$, or if PRNG splitting drifts, the host and TPU could execute different branches on the same step. If this happens:
1. The host logs the high-noise LR and keeps `loss_high` in the metric dictionary.
2. The device executes the low-noise branch, leaving `branch_metrics["loss_high"]` as `NaN`.
3. TensorBoard receives `NaN` for `learning/loss_high`, and `host_step_*` permanently drifts from `TrainState.step`.

**Recommendation**: Determine `is_high_noise` on the host and pass `is_high_noise: bool` (or `jnp.bool_(is_high_noise)`) directly as an argument into `p_train_step`. This guarantees 100% synchronization by construction and eliminates `cond_rng` from `p_train_step` without incurring recompilation in `jax.lax.cond`.

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,764 @@
...
+    - Homogeneous high batch: skips low-noise expert entirely (0 FLOPs for low expert).
+    - Mixed-timestep batch: executes at most one batched call per expert on the batch and
+      selects per-sample outputs via `jnp.where(is_high_mask, pred_high, pred_low)`.
+    This avoids unrolling $B$ single-sample model calls while guaranteeing exact per-sample
+    expert routing semantics and constant HLO graph complexity.
+  """
+  num_train_timesteps = scheduler.config.num_train_timesteps
+  boundary = int(config.boundary_ratio * num_train_timesteps)
+
+  latents = data["latents"].astype(config.weights_dtype)
```
mbohlool, line 714: 🟡 **Important**: Disconnect between `eval_step_2_2` and `test_eval_microbatching`.

[`test_eval_microbatching`](file:///mnt/workspace/maxdiffusion/src/maxdiffusion/tests/wan/wan_trainer_2_2_test.py#L558-L609) states in its docstring:
> *"Verify eval_step_2_2 microbatches when global_batch_size_to_train_on is smaller than loaded batch."*

However, `eval_step_2_2` does not reference `config.global_batch_size_to_train_on` or perform chunking; it passes the entire loaded `data["latents"]` batch into the models. If users set `global_batch_size_to_load` larger than `global_batch_size_to_train_on` (a common strategy to saturate I/O), this will feed the full loaded batch into the ~27B model during evaluation, triggering TPU OOM.

**Recommendation**:
- If batch-level execution without loop unrolling was chosen for constant HLO complexity, slice the evaluation data to `config.global_batch_size_to_train_on` (matching `train_step_2_2:L538`):
  ```python
  data = {k: v[: config.global_batch_size_to_train_on] for k, v in data.items()}
  ```
  and update the test docstring/assertions in `wan_trainer_2_2_test.py:L558` to reflect slicing rather than microbatching.
- Alternatively, if microbatching the loaded batch is intended, implement the chunking loop as done in [`wan_trainer.py:L231-L240`](file:///mnt/workspace/maxdiffusion/src/maxdiffusion/trainers/wan_trainer.py#L231-L240).

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,764 @@
...
+        next_batch_future = executor.submit(load_next_batch, train_data_iterator, example_batch, self.config)
+
+        # Track active expert on host without device-to-host sync stall
+        _, _, _, _, cond_rng, _ = jax.random.split(step_rng, num=6)
+        shift = getattr(pipeline.scheduler.config, "shift", 1.0) or 1.0
+        u_boundary = compute_unshifted_boundary(self.config.boundary_ratio, shift)
+        is_high_noise = bool(jax.random.uniform(cond_rng) >= u_boundary)
+        if is_high_noise:
+          active_lr = float(learning_rate_scheduler_high(host_step_high))
+          host_step_high += 1
```
mbohlool: 🟡 **Important**: Expert step counters are incremented before step logging.

`host_step_high += 1` and `host_step_low += 1` are incremented here before line 424, where `train_metric["steps"]` is assigned. As a result, the very first step executed by an expert (0-indexed step 0) is recorded in TensorBoard at step `1`. Step `0` is skipped for per-expert loss and learning rate curves, while global `step` starts at `0`.

**Recommendation**: Record `train_metric["steps"]` with the current step count, and increment `host_step_high` / `host_step_low` at the end of the training loop iteration.

src/maxdiffusion/pipelines/wan/wan_pipeline.py
```diff
@@ -317,7 +317,17 @@ def create_model(rngs: nnx.Rngs, wan_config: dict):
 
   # 1. Load config.
   if restored_checkpoint:
-    wan_config = restored_checkpoint["wan_config"]
+    if subfolder == "transformer" and "wan_config_high" in restored_checkpoint:
+      wan_config = dict(restored_checkpoint["wan_config_high"])
+    elif subfolder == "transformer" and hasattr(restored_checkpoint, "wan_config_high"):
```
mbohlool: 🟢 **Nit**: Guard against `None` value when `hasattr(restored_checkpoint, "wan_config_high")` is True.

If `restored_checkpoint` is an object with attribute `wan_config_high = None`, `dict(getattr(restored_checkpoint, "wan_config_high"))` will raise `TypeError: 'NoneType' object is not iterable`.

Consider:
```suggestion
    wan_config_high = (
        restored_checkpoint.get("wan_config_high")
        if isinstance(restored_checkpoint, dict)
        else getattr(restored_checkpoint, "wan_config_high", None)
    )
    if subfolder == "transformer" and wan_config_high is not None:
      wan_config = dict(wan_config_high)
    else:
```

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,764 @@
...
+  def _eval_low(operands):
+    l, t, e = operands
+    model = nnx.merge(state_low.graphdef, state_low.params, state_low.rest_of_state)
+    return model(hidden_states=l, timestep=t, encoder_hidden_states=e, deterministic=True)
+
+  def _skip_low(operands):
+    l, _, _ = operands
+    return jnp.zeros_like(l)
+
+  pred_high = jax.lax.cond(
```
mbohlool, line 742: 💡 **Suggestion**: When an evaluation batch contains mixed timesteps ($t \ge \text{boundary}$ and $t < \text{boundary}$), both `_eval_high` and `_eval_low` execute on the entire batch size $B$, resulting in $2B$ forward passes followed by `jnp.where`.

If the evaluation dataset iterator can group or sort samples by timestep into homogeneous buckets (either all high or all low), `jax.lax.cond` will bypass one expert entirely, halving the evaluation FLOPs and peak activation memory.

src/maxdiffusion/trainers/wan_trainer_2_2.py
```diff
@@ -0,0 +1,764 @@
...
+    noise_scheduler = FlaxFlowMatchScheduler(shift=shift, dtype=jnp.float32)
+    noise_scheduler_state = noise_scheduler.create_state()
+    noise_scheduler_state = noise_scheduler.set_timesteps(
+        noise_scheduler_state, num_inference_steps=1000, training=True, shift=shift
+    )
+    return noise_scheduler, noise_scheduler_state
+
+  def get_data_shardings(self, mesh):
+    data_sharding = jax.sharding.NamedSharding(mesh, P(*self.config.data_sharding))
+    data_sharding = {"latents": data_sharding, "encoder_hidden_states": data_sharding}
```
mbohlool:
```suggestion
    sharding = jax.sharding.NamedSharding(mesh, P(*self.config.data_sharding))
    return {"latents": sharding, "encoder_hidden_states": sharding}
```

mbohlool (changes requested):

src/maxdiffusion/trainers/wan_trainer_2_2.py
Toshi-31: Done

src/maxdiffusion/trainers/wan_trainer_2_2.py
Toshi-31: Done

src/maxdiffusion/trainers/wan_trainer_2_2.py
Toshi-31: eval_step_2_2 is already structured to skip the inactive expert for homogeneous batches, and bucketing in the dataset iterator can be explored as a future optimization.

mbohlool (approved):

PR #473, opened 2026-09-04
fix: prevent silent double-application of LoRA weights

WangXukang-cypher:
## Summary

- Wire up the previously dead `num_fused_loras` counter in `LoRABaseMixin` to track merged LoRA identities
- Before each merge, the loader checks whether the same `(path, weight_name)` was already applied and skips with a warning
- Covers all three NNX loaders: `Wan2_1NNXLoraLoader`, `Wan2_2NNXLoraLoader`, `LTX2NNXLoraLoader`

## Problem

`merge_lora` unconditionally adds `delta` to model weights (`kernel += delta`). Calling `load_lora_weights` twice with the same LoRA — via duplicate config entries or notebook cell re-execution — silently doubles the LoRA effect. The existing `num_fused_loras = 0` counter in `LoRABaseMixin` was never incremented or checked.

## Test plan

- [x] Pre-commit passes (ruff, pyink, pylint ≥ 7)
- [x] Verified: first merge allowed, duplicate blocked, different LoRA allowed, different weight name allowed, independent loader instances isolated, class-level counter unchanged
- [ ] Existing CI tests (require TPU environment)

### DIFF
--- a/src/maxdiffusion/loaders/lora_base.py
+++ b/src/maxdiffusion/loaders/lora_base.py
@@ -12,28 +12,51 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from ..models.modeling_utils import load_state_dict
-from ..utils import _get_model_file
+import os
 
 import safetensors
+
+from ..models.modeling_utils import load_state_dict
+from ..utils import _get_model_file
 
 
 class LoRABaseMixin:
   """Utility class for handing LoRAs"""
 
   _lora_lodable_modules = []
-  num_fused_loras = 0
 
-  def __init__(self):
-    self._fused_lora_keys = set()
+  @staticmethod
+  def _lora_key(lora_model_path, weight_name, target=None):
+    """Build a hashable key for a LoRA source and merge target."""
+    if isinstance(lora_model_path, dict):
+      source_key = ("state_dict", id(lora_model_path))
+    elif isinstance(lora_model_path, (str, os.PathLike)):
+      source_key = os.path.normpath(os.fspath(lora_model_path))
+    else:
+      try:
+        hash(lora_model_path)
+        source_key = lora_model_path
+      except TypeError:
+        source_key = (type(lora_model_path).__name__, id(lora_model_path))
 
-  def _check_and_record_lora(self, lora_key):
-    """Return True if this LoRA was already merged (duplicate). Records it otherwise."""
-    if lora_key in self._fused_lora_keys:
-      return True
-    self._fused_lora_keys.add(lora_key)
-    self.num_fused_loras += 1
-    return False
+    return (source_key, weight_name, target)
+
+  @staticmethod
+  def _is_lora_fused(pipeline, lora_key):
+    """Return whether a LoRA has already been merged into this pipeline."""
+    fused_lora_keys = getattr(pipeline, "_fused_lora_keys", None)
+    return fused_lora_keys is not None and lora_key in fused_lora_keys
+
+  @staticmethod
+  def _record_lora_fused(pipeline, lora_key):
+    """Record a successful merge on the pipeline that owns the mutated weights."""
+    fused_lora_keys = getattr(pipeline, "_fused_lora_keys", None)
+    if fused_lora_keys is None:
+      fused_lora_keys = set()
+      pipeline._fused_lora_keys = fused_lora_keys
+    if lora_key not in fused_lora_keys:
+      fused_lora_keys.add(lora_key)
+      pipeline.num_fused_loras = getattr(pipeline, "num_fused_loras", 0) + 1
 
   def load_lora_weights(self, **kwargs):
     raise NotImplementedError("`load_lora_weights()` is not implemented.")
@@ -57,6 +80,7 @@ class LoRABaseMixin:
   ):
     from .lora_pipeline import LORA_WEIGHT_NAME_SAFE
 
+    state_dict = pretrained_model_name_or_path_or_dict
     model_file = None
     if not isinstance(pretrained_model_name_or_path_or_dict, dict):
       # Let's first try to load .safetensors weights
--- a/src/maxdiffusion/loaders/lora_base.py
+++ b/src/maxdiffusion/loaders/lora_base.py
@@ -24,6 +24,17 @@ class LoRABaseMixin:
   _lora_lodable_modules = []
   num_fused_loras = 0
 
+  def __init__(self):
+    self._fused_lora_keys = set()
+
+  def _check_and_record_lora(self, lora_key):
+    """Return True if this LoRA was already merged (duplicate). Records it otherwise."""
+    if lora_key in self._fused_lora_keys:
+      return True
+    self._fused_lora_keys.add(lora_key)
+    self.num_fused_loras += 1
+    return False
+
   def load_lora_weights(self, **kwargs):
     raise NotImplementedError("`load_lora_weights()` is not implemented.")
 
--- a/src/maxdiffusion/loaders/ltx2_lora_nnx_loader.py
+++ b/src/maxdiffusion/loaders/ltx2_lora_nnx_loader.py
@@ -54,9 +54,14 @@ class LTX2NNXLoraLoader(LoRABaseMixin):
       max_logging.log("No LoRA weight name provided; skipping LoRA load.")
       return pipeline
 
-    lora_key = (lora_model_path, transformer_weight_name)
-    if self._check_and_record_lora(lora_key):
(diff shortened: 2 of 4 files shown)


src/maxdiffusion/loaders/lora_base.py
```diff
@@ -24,6 +24,17 @@ class LoRABaseMixin:
   _lora_lodable_modules = []
   num_fused_loras = 0
 
+  def __init__(self):
+    self._fused_lora_keys = set()
```
Perseus14: Defining an `__init__` in a mixin class can be tricky in Python because mixins participate in multiple inheritance. If a subclass doesn't explicitly call `super().__init__()`, `self._fused_lora_keys` won't be initialized and might raise an `AttributeError`.

More importantly, since users often instantiate loaders on the fly (e.g., `lora_loader = Wan2_1NNXLoraLoader()`), storing the set on `self` means creating a new loader in a notebook cell will reset the set to empty and won't guard against cell re-runs.

A clean alternative is to store the tracking set on the `pipeline` instance being modified (for example, `getattr(pipeline, "_fused_lora_keys", set())`). That way, the tracked state always travels with the actual weights being mutated.

src/maxdiffusion/loaders/lora_base.py
```diff
@@ -24,6 +24,17 @@ class LoRABaseMixin:
...
+
+  def _check_and_record_lora(self, lora_key):
+    """Return True if this LoRA was already merged (duplicate). Records it otherwise."""
+    if lora_key in self._fused_lora_keys:
+      return True
+    self._fused_lora_keys.add(lora_key)
+    self.num_fused_loras += 1
+    return False
```
Perseus14: It’s safer to split the "check" and "record" steps into two separate methods here:

1. **Check first:** Before loading, check if the LoRA is already merged. If so, log and return early.
2. **Record after success:** Only add to the set and increment `num_fused_loras` *after* the weights are successfully loaded and merged via `merge_fn`.

If we record beforehand and `lora_state_dict()` raises an exception (like a network timeout or file not found), the key remains marked as fused. If the user catches the error or retries, it would be skipped on the second attempt.

Also, keep in mind that `lora_model_path` can sometimes be a `dict` (in-memory weights). Since dictionaries are unhashable, adding a tuple containing a dict to a `set` will raise a `TypeError`. We could fall back to `id(lora_model_path)` or a custom string tag if it's a dict.

src/maxdiffusion/loaders/wan_lora_nnx_loader.py
```diff
@@ -50,6 +50,11 @@ def load_lora_weights(
     def translate_fn(nnx_path_str):
       return lora_conversion_utils.translate_wan_nnx_path_to_diffusers_lora(nnx_path_str, scan_layers=scan_layers)
 
+    lora_key = (lora_model_path, transformer_weight_name)
+    if self._check_and_record_lora(lora_key):
+      max_logging.log(f"WARNING: LoRA '{lora_model_path}' already merged — skipping to avoid double-application.")
+      return pipeline
+
```
Perseus14: We should move this check inside the `if hasattr(pipeline, "transformer") and transformer_weight_name:` block.

If `transformer_weight_name` is empty or `pipeline` doesn't have a transformer, we don't want to record the key or increment the fused count since no weights were actually modified.

Minor tip: It's also helpful to run `os.path.normpath` or `os.path.abspath` on `lora_model_path` (when it's a string path) so that `./checkpoints/my_lora` and `checkpoints/my_lora` are recognized as the same file.

src/maxdiffusion/loaders/wan_lora_nnx_loader.py
```diff
@@ -91,7 +96,10 @@ def translate_fn(nnx_path_str: str):
       return lora_conversion_utils.translate_wan_nnx_path_to_diffusers_lora(nnx_path_str, scan_layers=scan_layers)
 
     # Handle high noise model
-    if hasattr(pipeline, "high_noise_transformer") and high_noise_weight_name:
+    high_key = (lora_model_path, high_noise_weight_name, "high_noise")
+    if self._check_and_record_lora(high_key):
+      max_logging.log(f"WARNING: LoRA '{lora_model_path}' already merged into high_noise_transformer — skipping.")
+    elif hasattr(pipeline, "high_noise_transformer") and high_noise_weight_name:
```
Perseus14: Check the order of conditions here: if `high_noise_weight_name` is `None` (for instance, if a user only wants to condition the low-noise stage), `high_key` will be `(path, None, "high_noise")`. Calling `_check_and_record_lora` will record this key and increment `num_fused_loras`, and then fall into the `else` branch without merging anything.

If we nest the duplicate check inside the `if hasattr(...) and weight_name:` branch, we'll ensure we only check and record when there are valid weights to apply.

src/maxdiffusion/loaders/wan_lora_nnx_loader.py
```diff
@@ -100,7 +108,10 @@ def translate_fn(nnx_path_str: str):
       max_logging.log("high_noise_transformer not found or no weight name provided for LoRA.")
 
     # Handle low noise model
-    if hasattr(pipeline, "low_noise_transformer") and low_noise_weight_name:
+    low_key = (lora_model_path, low_noise_weight_name, "low_noise")
+    if self._check_and_record_lora(low_key):
+      max_logging.log(f"WARNING: LoRA '{lora_model_path}' already merged into low_noise_transformer — skipping.")
+    elif hasattr(pipeline, "low_noise_transformer") and low_noise_weight_name:
```
Perseus14: Same as above

src/maxdiffusion/loaders/ltx2_lora_nnx_loader.py
```diff
@@ -54,6 +54,11 @@ def translate_fn(nnx_path_str):
       max_logging.log("No LoRA weight name provided; skipping LoRA load.")
       return pipeline
 
+    lora_key = (lora_model_path, transformer_weight_name)
+    if self._check_and_record_lora(lora_key):
+      max_logging.log(f"WARNING: LoRA '{lora_model_path}' already merged — skipping to avoid double-application.")
+      return pipeline
+
```
Perseus14: Similar to Wan, let's defer recording the key until after the weights have been applied to `pipeline.transformer` and/or `pipeline.connectors`. That keeps the tracking state accurate even if a file load fails midway.

Perseus14 (changes requested):
Thanks for tackling this, @WangXukang-cypher!

The general direction makes a lot of sense, but there are a few edge cases around state ownership, error handling, and Python mixin design that we should address before merging:

1. **Where state should live (Pipeline vs. Loader):** In MaxDiffusion, loaders like `Wan2_1NNXLoraLoader` are typically instantiated as transient helpers (e.g. `Wan2_1NNXLoraLoader().load_lora_weights(...)`). If a user re-runs a notebook cell that creates a new loader instance, `self._fused_lora_keys` resets and the weights can still be doubly merged. Conversely, sharing one loader across different pipelines might inadvertently block the second pipeline from loading its LoRA. Attaching the tracked keys to the `pipeline` (or module) instead of the loader solves both issues cleanly.
2. **Atomic registration:** Currently, `_check_and_record_lora` records the LoRA as merged before we know whether downloading/reading the weights succeeds or if the model has matching layers. Splitting this into a pre-check (to skip if already present) and a post-merge record (only after `merge_fn` finishes) makes the loader resilient to network glitches and file errors.
3. **Input handling:** In maxdiffusion, LoRA sources can sometimes be passed as in-memory state dicts (`dict`), which aren't hashable in a `set`. We should handle this case safely.
4. **Tests:** Could you add a couple of unit tests in `src/maxdiffusion/tests/` to verify duplicate skipping and ensure retry/fresh pipeline behavior works as expected?

I’ve left some inline suggestions with code snippets to help guide the refactor. Happy to discuss further!

WangXukang-cypher:
Thanks for the detailed review, @Perseus14! I've updated the PR in `573fe4c` based on your suggestions:

- Moved fused-LoRA tracking from transient loader instances to the pipeline that owns the mutated weights. This preserves duplicate protection across newly created loaders while keeping separate pipelines independent.
- Split duplicate checking from registration and now record a LoRA only after its load and merge complete successfully.
- Moved the checks inside the valid-target branches, so missing transformers or weight names are not recorded.
- Normalized path-based keys and added safe identity-based keys for in-memory state dictionaries.
- Applied the same post-merge registration behavior to WAN 2.1, WAN 2.2, and LTX2.
- Added unit tests covering duplicate skipping across loader instances, fresh-pipeline behavior, retry after load/merge failures, in-memory dictionaries, and missing WAN 2.2 weight names.

Thanks again for pointing out these edge cases. Please take another look when you have a chance!

PR #477, opened 2026-09-13
feat(attention): fixed-m splash attention kernel with dynamic bounds and safety fallbacks

Perseus14:
### Summary

Implements the single-device fixed-m splash attention kernel using Pallas on Cloud TPU.

Instead of tracking an online-softmax running max per KV block, eligible (head, Q-block) pairs
subtract a *precomputed* shift derived from a Cauchy-Schwarz bound on the logits, which lets the
numerator and denominator accumulate directly in FP32.

- **Dynamic eligibility bounding.** The two constants are derived from the actual KV length rather
  than pinned to one operating point:
  $$C(N) = 127 - \lceil\log_2 N\rceil - \lceil\log_2 V_{\max}\rceil, \qquad W(N) = C(N) + 125, \qquad \text{gate} = \left\lfloor \frac{W(N)}{2} \right\rfloor$$
  With the shift $m_i = \lceil U_i \rceil - C(N)$ built from the Cauchy-Schwarz bound
  $U_i = \max_i \lVert q_i \rVert \max_j \lVert k_j \rVert$, every shifted exponent satisfies
  $z_j - m_i \ge C(N) - (U_i + \lceil U_i \rceil) \ge -125 > -126$, so no term flushes to subnormal
  and no accumulator overflows.
- **Virtual K-centering, opt-in.** Centers logits via the register-level projection
  $q_i^\top \bar{k}$ without ever writing $(K - \bar{k})$ back to HBM. This shrinks $U_i$ (and so
  admits more heads); it does **not** loosen the gate — see below. On the ring path it activates
  **only when the caller supplies `k_mean`**; see the next section for why that is load-bearing.
- **Ragged sequence handling.** Dynamic tail slicing in `last_compute_body_fixed` lets physically
  unpadded KV sequences run without out-of-bounds access.
- **Relayout optimization.** Hoists the base-2 logit rescale $Q \cdot \log_2 e$ above the Ulysses
  all-to-all so XLA fuses the scalar multiply into the upstream projection, saving 185 MB of HBM
  traffic per layer.
- **Input validation & API hygiene.** Explicitly rejects legacy 2D `mk` arrays (`[H, num_q_blocks]`) at entry with a clear `ValueError`, preserves keyword-only `*,` delimiter in `make_custom_ring_attention`, and derives head-dim padding dynamically from `query.shape[-1]`.

### This PR is additive: `main`'s $R > 1$ behaviour is unchanged

Centering and the Cauchy-Schwarz bound are a **matched pair**. If the kernel exponentiates centered
logits $q \cdot (k_j - \bar{k})$ while the caller's eligibility bound was built from raw $K$, the
bound caps the wrong quantity — $\lVert k \rVert$ does not bound $\lVert k - \bar{k} \rVert$ — and a
tile can clear the gate while the logit it was supposed to cap overflows fp32.
`RingRawKeyBoundUnsoundTest` pins that down in pure arithmetic: keys all of norm 100 give a raw bound
of 100 against a gate of 116, while the centered max logit is ~200 and the shifted exponent exceeds
$2^{128}$.

This revision makes centering **opt-in**: the kernel centers only when handed a `k_mean`.
Callers that supply one also supply centered norms; callers that do not keep the uncentered path,
which is sound against the two-sided $\lfloor W/2 \rfloor$ bound. `main`'s $R > 1$ caller passes no
`k_mean`, so it is left running exactly as before and the restriction is gone. Net effect: this PR
adds the fixed-m kernel without changing any behaviour `main` already has, and **merges safely on its
own**. #478 then turns centering on by supplying both halves of the pair.

### Why the gate is halved, and why `is_ring` does not change it

Review raised that `get_fixed_m_constants` accepts `is_ring` but ignores it, and that the gate is
half the legacy `_FIXED_M_SAFE_BOUND = 213`. Both observations are correct; the conclusion that the
centered path should get the full window $W$ is not.

The appealing argument is that K-centering forces the realized row max $M \ge 0$, so only one side
needs absorbing. It fails because the shift is built from $U$, not from $M$: the worst case stays
$C - (U + \lceil U \rceil)$ whether or not the keys are centered. Wiring `is_ring` up to return $W$
for the Ulysses path reproduces the exact bug the halving was introduced to fix — at $N = 4096$,
$W = 232$, so a query of norm $231.01$ against exactly-centered keys passes the gate and then
silently loses **11.8% of the softmax mass** (0.882 vs 1.0) as its negative logits flush to zero.
`test_adversarial_centered_keys_softmax_mass_loss` and `test_cpu_proof_invariant_bounds` both pin
this down.

**Measured cost of the strict gate.** Instrumented over a full WAN 2.2 720p generation on v6e-8
(1280 metadata calls per gate, $N = 75{,}600$):

| gate | derivation | entries falling back | calls with any fallback |
|---:|---|---:|---:|
| **113** | $\lfloor W/2 \rfloor$ (shipped) | **0.73%** | 112 / 1280 |
| 201 | relative mass loss $\le 2^{-10}$ incl. the $N$ multiplicity | 0.45% | 56 / 1280 |
| 218 | relative mass loss $\le 2^{-10}$ per term | 0.44% | 56 / 1280 |
| 227 | $W$ | 0.43% | 54 / 1280 |

So the strict gate costs about 0.3 percentage points of extra fallback over the loosest alternative.

### Verification

- `custom_splash_fixed_m_test.py`: **23 passed** on TPU (dynamic bounds, non-divisible sequences, legacy 2D `mk` rejection, adversarial centered-key regression, hybrid fallbacks).
- `ring_fixed_m_test.py`: **36 passed, 1 skipped** on TPU.
- Rebased cleanly on latest `origin/main` (`1bc54811`).
- Linting: `pyink --pyink-indentation=2 --line-length=125` and `ruff check` clean.


src/maxdiffusion/kernels/custom_splash_attention.py
```diff
@@ -105,26 +183,31 @@ def _flash_attention_kernel(
...
+    # Virtual K-centering applies the row-specific projection: m_i = m_B + q_i^T \bar{k}.
+    m_base = mk_ref[0, h, i]
+    if k_mean_ref is not None:
+      qf = q_ref[...].astype(float32)
+      km = k_mean_ref[h // q_heads_per_kv_head, :].astype(float32)
+      mu = (qf * km[None, :]).sum(axis=1)[None, :]
+      m_fixed = m_base + mu
+    else:
+      m_fixed = m_base
     m_scratch_ref[...] = jnp.broadcast_to(m_fixed, m_scratch_ref.shape)
```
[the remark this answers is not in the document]
Perseus14: Done!

syhuang22 (commented):
A few things before this can go in though:

- If this lands on its own, fixed-m gets turned off for R>1 (`use_fixed_m = False` in `_ulysses_ring_custom_attention`, it only comes back in #478). That breaks the recipe we ship today. Can we split the stack so each PR is safe to merge by itself?
- Cutting the Ulysses gate from 213 to 113 feels more conservative than we need. With centered keys, the mass you can lose is bounded by 2^(ceil(U)-C-126), so a gate around C+116 already keeps the loss under 0.1%. Could you share fallback rates so we can pick the number?
- `mk[0]` means something different now, but a 2D `mk` still gets silently broadcast. Can we just raise on that?
- The V check never fired in my runs on WAN 2.2 (every call passed), and it only exists because C moved from 88 to 102. Is it worth the extra complexity?
- Please put back the comments that explain the Mosaic cliff and k-smoothing. They save the next person a lot of pain.

src/maxdiffusion/models/attention_flax.py
```diff
@@ -1356,6 +1454,24 @@ def _ulysses_ring_custom_attention(
...
+  # relies on. Centering the norms requires the same global k_mean the kernel
+  # computes internally, which is not plumbed to the caller in this change.
+  # R > 1 therefore falls back to online softmax, which needs no bound at all.
+  if use_fixed_m and num_ring_shards > 1:
+    max_logging.log(
+        f"[attention] fixed-m is not available for ring degree R={num_ring_shards} "
+        "(R = context_shards // ulysses_shards); falling back to online softmax. "
+        "Set ulysses_shards == ici_context_parallelism for R=1 to use fixed-m."
+    )
+    use_fixed_m = False
```
syhuang22: This turns off fixed-m for R>1, which is the ring2 x uly2 recipe we ship today. #478 turns it back on, so let's not land this one alone.

src/maxdiffusion/kernels/custom_splash_attention.py
```diff
@@ -47,30 +48,104 @@ def __init__(
...
+  # Safe window W(N) = C(N) - (-126.0) - 1.0 = C(N) + 125.0
+  safe_window = recenter - fp32_min_normal_exp - 1.0
+
+  # For integer threshold K = floor(W / 2), U <= K guarantees U + ceil(U) <= 2K <= W(N).
+  # For ANY logit S in [-U, U], the shifted exponent S - m_base satisfies:
+  #   S - m_base >= -U - (ceil(U) - C(N)) = C(N) - (U + ceil(U)) >= C(N) - W(N) = -125.0 >= -126.0.
+  # This guarantees that negative logits never flush to zero in normal FP32, preventing silent
+  # loss of significant softmax probability mass even when keys are mean-centered.
+  # Both ring and non-ring paths strictly adhere to this two-sided bound.
+  safe_bound = float(int(safe_window // 2))
```
syhuang22, line 189: This halves the Ulysses gate (213 -> 113). With centered keys the row mean is 0, so Jensen caps the lost mass at 2^(ceil(U)-C-126). Something like C+116 keeps it under 2^-10 without halving. What do fallback rates look like at 113?

src/maxdiffusion/kernels/custom_splash_attention.py
```diff
@@ -47,30 +48,104 @@ def __init__(
...
+
+  Returns:
+    True if `2**recenter` is representable in `dtype`.
+  """
+  return float(jnp.finfo(jnp.dtype(dtype)).maxexp) > float(recenter)
+
+
+def get_fixed_m_constants(
+    kv_seq_len: int,
+    is_ring: bool = False,
```
syhuang22, line 116: `is_ring` isn't used. Drop it?

src/maxdiffusion/kernels/custom_splash_attention.py
```diff
@@ -499,7 +591,37 @@ def _splash_attention_forward(
...
+    fixed_m_recenter, _ = get_fixed_m_constants(actual_kv_seq_len, is_ring=False)
+
+  # Scalar-prefetch operand carrying per-head / per-Q-block fixed-m data:
+  #   mk[0, h, i] = m_B (precomputed block fixed-m base shift), mk[1, h, i] = eligibility.
+  # A dummy is supplied for online callers; the kernel ignores it.
+  if use_fixed_m and mk is None:
+    raise ValueError("`mk` metadata array is required when `use_fixed_m=True`.")
+  if mk is None:
+    mk = jnp.zeros((2, num_q_heads, grid_height), jnp.float32)
+  elif mk.ndim == 2:
```
syhuang22, line 780: `mk[0]` used to be max||k||, now it's m. Broadcasting an old 2D `mk` would silently read one as the other. Can we just raise? (same at L797)

src/maxdiffusion/kernels/custom_splash_attention.py
```diff
@@ -105,26 +191,32 @@ def _flash_attention_kernel(
   # a single body per block, and the fixed bound stays PINNED through the
   # ragged last KV block, so every hop reports the identical m and the hops
   # combine by plain accumulation.
-  #
-  # Both must move together. Pinning without the uniform promise needs a
-  # SECOND body in the last block (fixed and online), and a two-body last
-  # block degrades the instruction schedule of the WHOLE grid -- measured 3x
-  # slower end to end, which is the cliff the design doc's D3 warns about.
-  # Keeping this one flag rather than two makes that combination unspellable.
   fixed_only = use_fixed_m and uniform_fixed_m
```
syhuang22, line 251: Can we keep the comment that was here? It's the only thing explaining why pinning and `uniform_fixed_m` must go together (two-body last block = Mosaic cliff, ~3x slower).

syhuang22 (commented):
Some line-level notes to go with my comment above.

src/maxdiffusion/kernels/custom_splash_attention.py
```diff
@@ -642,12 +786,16 @@ def v_index_map(h, i, j, *_):
   grid_height = (actual_q_seq_len + bq - 1) // bq
   grid = (num_q_heads, grid_height, grid_width)
 
-  # Scalar-prefetch operand carrying per-head fixed-m data (same convention as
-  # `_splash_attention_forward`): mk[0, h] = max_j||k_j|| over ALL ring shards
-  # (the caller all-reduces this over the ring axis), mk[1, h] = eligibility.
+  # Scalar-prefetch operand carrying per-head / per-Q-block fixed-m data:
+  #   mk[0, h, i] = max_j||k_j|| over ALL ring shards
```
syhuang22: Nit: comment still says `mk[0]` is max||k||.

src/maxdiffusion/kernels/custom_splash_attention.py
```diff
@@ -47,30 +48,104 @@ def __init__(
...
-# Fixed-m softmax-bound constants. Instead of tracking the online-softmax
-# running max per KV block, eligible heads subtract a precomputed per-query
-# upper bound on the logits (Cauchy-Schwarz: max_j q_i.k_j <= ||q_i|| *
-# max_j||k_j||). _FIXED_M_RECENTER (C) shifts the exp2 exponents up so the
-# largest surviving term stays above the f32 subnormal-flush floor 2^-126:
-# with k-smoothing the per-row max is >= 0, so the max term has exponent
-# >= -ceil(bound) + C, which stays > -126 while ceil(bound) <=
-# _FIXED_M_SAFE_BOUND (= C + 126 - 1 of margin). Heads whose worst-case bound
-# exceeds the gate fall back to online softmax (the "sink" heads).
+# Legacy fixed-m bounds preserved for backwards compatibility
```
syhuang22: Only the tests use these now. Delete, or say why they're kept?

src/maxdiffusion/models/attention_flax.py
```diff
@@ -840,6 +840,69 @@ def ring_scan_body(carry, _):
...
+    if num_q_heads % num_kv_heads != 0:
+      raise ValueError(f"num_q_heads ({num_q_heads}) must be divisible by num_kv_heads ({num_kv_heads}) for GQA fixed-m.")
+    q_heads_per_kv_head = num_q_heads // num_kv_heads
+    mk_h_sq = jnp.repeat(mk_h_sq, q_heads_per_kv_head, axis=1)  # (batch, num_q_heads)
+
+  dtype_safe = custom_splash.fixed_m_dtype_is_safe(query.dtype, recenter)
+  v_ok = 1.0 if dtype_safe else 0.0
+  if dtype_safe and value is not None:
+    v_max_sq = (value.astype(jnp.float32) ** 2).max()
+    v_ok = (v_max_sq <= (v_max_bound**2)).astype(jnp.float32)
```
syhuang22, line 880: This never tripped in my WAN 2.2 runs. It's only needed because C went 88 -> 102, which buys gate 106 -> 113. Worth it?

src/maxdiffusion/models/attention_flax.py
```diff
@@ -923,27 +980,39 @@ def wrap_ulysses_attention(query, key, value, attention_mask):
...
+      recenter, safe_bound = custom_splash.get_fixed_m_constants(actual_kv_seq_len, is_ring=False)
+
+      k_mean = None
       if use_fixed_m:
-        # k-smoothing (output-invariant): subtracting the per-row key mean
-        # forces every logit row to have mean 0, hence row-max >= 0 — the
-        # precondition that keeps the fixed-m Cauchy-Schwarz bound flush-free.
-        key = key - jnp.mean(key, axis=2, keepdims=True)
+        k_mean = jnp.mean(real_key.astype(jnp.float32), axis=2)
+        if k_mean.shape[-1] < 128:
```
syhuang22, line 996: Nit: hard-coded 128, can we use the padded head dim? (same at L1556)

src/maxdiffusion/models/attention_flax.py
```diff
@@ -1774,6 +1942,32 @@ def ulysses_custom_fixed_m_kernel(q, k, v, context):
       use_base2_exp=context.get("use_base2_exp", True),
       use_experimental_scheduler=context.get("use_experimental_scheduler", False),
       use_fixed_m=True,
+      per_q_block=False,
```
syhuang22, line 1927: Heads-up: `ulysses_custom_fixed_m` goes from per-row m to per-head m here. Fine for safety, but worth mentioning in the description.

src/maxdiffusion/kernels/splash_attention/ring_attention_kernel.py
```diff
@@ -1132,7 +1255,6 @@ def _lse_scan(_):
 
 
 def make_custom_ring_attention(
```
syhuang22, line 1261: Was removing the keyword-only `*` on purpose? With this many bool args it's a nice guard.

src/maxdiffusion/generate_ltx_video.py
```diff
@@ -213,9 +213,7 @@ def run(config):
 
   for prompt_idx, current_prompt in enumerate(prompts):
     prompt_word_count = len(current_prompt.split())
-    enhance_prompt = (
-        prompt_enhancement_words_threshold > 0 and prompt_word_count < prompt_enhancement_words_threshold
-    )
+    enhance_prompt = prompt_enhancement_words_threshold > 0 and prompt_word_count < prompt_enhancement_words_threshold
```
syhuang22, line 216: Unrelated formatting change, mind dropping it?

src/maxdiffusion/models/attention_flax.py
Perseus14: Fixed in `44dfe632` — removed this `use_fixed_m = False` override and made K-centering opt-in (`k_mean is not None`) on the ring path, so `R > 1` continues to run fixed-m safely when PR #477 is merged on its own.

src/maxdiffusion/kernels/custom_splash_attention.py
Perseus14: Good catch — updated both `_splash_attention_forward` (`L780`) and `_splash_attention_forward_ring` (`L981`) to raise an explicit `ValueError` when `mk.ndim == 2`, and added `test_legacy_2d_mk_raises_value_error` in `custom_splash_fixed_m_test.py`.

src/maxdiffusion/kernels/custom_splash_attention.py
Perseus14: Restored in `custom_splash_attention.py:L239-L250`!

src/maxdiffusion/kernels/custom_splash_attention.py
Perseus14: Fixed in `custom_splash_attention.py:L774` and `L973` — both comments now accurately document `mk[0, h, i] = m_B` (precomputed block fixed-m base shift).

src/maxdiffusion/kernels/custom_splash_attention.py
Perseus14: Added an explicit explanation at `custom_splash_attention.py:L67-L70` noting that these legacy constants are retained as the static reference point against which `get_fixed_m_constants(N)` is calibrated and tested in `custom_splash_fixed_m_test.py`.

src/maxdiffusion/models/attention_flax.py
Perseus14: Done! Moved `_pad_data_for_flash` before the `k_mean` pad in `_ulysses_attention` (`attention_flax.py:L1071`) and replaced hard-coded `128` with `query.shape[-1]` (matching `q.shape[-1]` in `ring_attention_kernel.py:L945`).

src/maxdiffusion/models/attention_flax.py
Perseus14: Good callout! Updated the PR description to explicitly note that `ulysses_custom_fixed_m` uses per-head $m_B$ metadata (`per_q_block=False`), while `ulysses_custom_fixed_m_per_q_block` provides finer per-Q-block granularity.

src/maxdiffusion/kernels/splash_attention/ring_attention_kernel.py
Perseus14: Restored `*,` after `orig_kv_seq_len: int` in `make_custom_ring_attention` (`ring_attention_kernel.py:L1308`). All callers already use keyword arguments.

src/maxdiffusion/generate_ltx_video.py
Perseus14: Dropped in `44dfe632` — `generate_ltx_video.py` is completely untouched now.

src/maxdiffusion/models/attention_flax.py
Perseus14: Yes — raising $C(N)$ from `88` to `102` gives us +7 on the safe bound (`106 -> 113`), which meaningfully improves fixed-m hit rates. On the ring path (`_ring_fixed_m_norms_pre_a2a`), `vn_local` is fused into the single `(ulysses, ring)` `pmax` collective alongside `qn` and `kn`, so it adds zero extra collectives and `<0.05 ms/step` while guaranteeing fail-closed protection against FP32 numerator overflow (`inf`/`NaN`) if a LoRA or fine-tuned checkpoint produces `|V| > 256`.

src/maxdiffusion/kernels/custom_splash_attention.py
Perseus14: Kept `is_ring: bool = False` in the signature (with `del is_ring` and an explicit note at L166–188) so the ~20 callers across `attention_flax.py`, `ring_attention_kernel.py`, and the test suite self-document whether they are passing the distributed ring sequence length (`shard_len * R`) vs local shard length. Happy to remove the keyword arg across those call sites if you prefer a stricter signature!

src/maxdiffusion/kernels/custom_splash_attention.py
Perseus14: On Wan 2.2 720p/81f (40 steps), `per_q_block=True` + K-centering shrinks the realized Cauchy-Schwarz product $U = \|q\| \cdot \|k - \bar{k}\|$ by $>2.2\times$, so **99.4% of Q-blocks** pass gate `113` on 1D Ulysses (`U=4`), and **82.0% of calls** pass on Ring `R=2` (higher than `main`'s 70.0% at gate `213`).

The reason we keep `floor(W/2) = 113` even with centered keys is that $m_B = \lceil U_{\text{block}} \rceil - C$ is shared across all rows in a Q-block (or head), not computed per individual row. If a block has an outlier row with $U_{\text{block}} \approx C + 116$ and a quiet row with $\|q_i\| \approx 0$ ($z_{i,j} \approx 0$), shifting the quiet row by $m_B$ gives exponents $0 - m_B \approx -116$. If $U_{\text{block}}$ approached $C + 126$, quiet rows in that block would underflow to zero denominator ($0/0 = \text{NaN}$) despite centered keys. `floor(W/2)` prevents both negative-logit underflow on high-norm rows and zero-logit underflow on quiet rows in the same block.

syhuang22 (dismissed):

src/maxdiffusion/models/attention_flax.py
```diff
@@ -840,6 +840,69 @@ def ring_scan_body(carry, _):
...
+    q_heads_per_kv_head = num_q_heads // num_kv_heads
+    mk_h_sq = jnp.repeat(mk_h_sq, q_heads_per_kv_head, axis=1)  # (batch, num_q_heads)
+
+  dtype_safe = custom_splash.fixed_m_dtype_is_safe(query.dtype, recenter)
+  v_ok = 1.0 if dtype_safe else 0.0
+  if dtype_safe and value is not None:
+    v_max_sq = (value.astype(jnp.float32) ** 2).max()
+    v_ok = (v_max_sq <= (v_max_bound**2)).astype(jnp.float32)
+
+  num_q_blocks = q_len // block_q
```
eltsai, line 895: nit: here we are using floor division but the kernel uses `grid_height = (actual_q_seq_len + bq - 1) // bq` (ceil). Can we do a `if q_len % block_q != 0` check here?

src/maxdiffusion/kernels/splash_attention/ring_attention_kernel.py
```diff
@@ -859,6 +864,12 @@ def _custom_ring_attention_forward(
     bidirectional: bool = False,
     use_fixed_m: bool = False,
     fixed_m_norms: tuple[jax.Array, jax.Array] | None = None,
+    fixed_m_norms_squared: bool = True,
+    per_q_block: bool = False,
```
eltsai: this is set as True by default in other places, we can change it for consistency

src/maxdiffusion/models/attention_flax.py
```diff
@@ -1469,23 +1589,51 @@ def wrap_ulysses_ring_attention(query, key, value):
...
+            fixed_m_recenter=recenter,
         )
+
+        def _run_uniform(q, k, v, m, km):
+          return jax.vmap(splash_kernel_uniform, in_axes=(0, 0, 0, 0, 0))(q, k, v, m, km)
+
+        def _run_hybrid(q, k, v, m, km):
+          return jax.vmap(splash_kernel_hybrid, in_axes=(0, 0, 0, 0, 0))(q, k, v, m, km)
+
+        raw_out = jax.lax.cond(all_fixed, _run_uniform, _run_hybrid, query, key, value, mk_arr, k_mean)
```
eltsai, line 1640: I wonder what is the cost of the lax.cond here (maybe it's worth separating a cond-free variant)

src/maxdiffusion/models/attention_flax.py
Perseus14: Thanks! `_compute_fixed_m_metadata` now raises a `ValueError` when `q_len % block_q != 0`. That way the per-block metadata can't silently cover fewer blocks than the kernel's ceil-div `grid_height`.

src/maxdiffusion/kernels/splash_attention/ring_attention_kernel.py
Perseus14: Done!

PR #484, opened 2026-09-16
Fix optimizer state restoration and multi-host sharding in WAN trainer

ninatu:
Previously, optimizer state restoration was silently failing as a no-op because `state.replace(opt_state=...)` returns a new TrainState rather than mutating in place, and its return value was discarded.

Fixing the assignment exposed that the restored optimizer state had no device sharding applied:

1. Restored optimizer state from checkpoints must be explicitly sharded across the device mesh (`state_shardings.opt_state`) to run on device.

2. Because Orbax checkpoint key paths diverge from live Optax PyTree structures (string indices vs. JAX KeyEntry objects and trailing `.value` wrappers), `restore_optimizer_state_by_path` was added to map restored checkpoint leaves onto the target sharded PyTree structure.

3. In `max_utils.device_put_replicated`, populating device shards from host arrays via `jax.make_array_from_callback` caused eager cross-process coordination failures across multi-host setups. Converting input arrays to local NumPy arrays (`np.asarray(arr)`) allows clean, process-local slicing per device shard without inter-host dispatch.


src/maxdiffusion/max_utils.py
```diff
@@ -503,9 +503,13 @@ def device_put_replicated(x, sharding):
...
   to also shard an array based on sharding.
   """
+  if isinstance(x, jax.Array) and hasattr(x, "sharding") and x.sharding == sharding:
+    return x
+
   arr = getattr(x, "value", x)
+  arr_np = np.asarray(arr)
   shd = getattr(sharding, "value", sharding)
-  res = jax.make_array_from_callback(arr.shape, shd, lambda index: arr[index])
+  res = jax.make_array_from_callback(arr_np.shape, shd, lambda index: arr_np[index])
```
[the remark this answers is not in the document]
ninatu: For Wan optimizer restoration, `jax.device_put(arr, shd)` does not work in multi-host setups and fails with:

`ValueError: When the second argument to device_put is a Device, the first argument must be a fully addressable array or a non-addressable array with a single device sharding. Got value with devices {CpuDevice(id=4096), CpuDevice(id=6144), CpuDevice(id=2048), CpuDevice(id=0)}`

Additionally, **the array is already on the CPU host**: the restored checkpoint leaves are loaded from Orbax on a replicated CPU mesh, not on accelerator devices, so `np.asarray(arr)` is not performing an expensive device-to-host transfer.

src/maxdiffusion/trainers/base_wan_trainer.py
```diff
@@ -49,6 +52,59 @@ def _to_array(x):
...
+    target_opt_shardings: Any,
+    restored_opt_state: Any,
+) -> Any:
+  """Restores optimizer state by matching exact parameter paths and applying sharding."""
+  if isinstance(restored_opt_state, (dict, flax.core.FrozenDict)):
+    flat_restored = flax.traverse_util.flatten_dict(restored_opt_state)
+    raw_items = flat_restored.items()
+  else:
+    leaves_with_path, _ = jax.tree_util.tree_flatten_with_path(restored_opt_state)
+    raw_items = leaves_with_path
```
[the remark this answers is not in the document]
ninatu: Done.

entrpn:
@Perseus14 can you take a look?

src/maxdiffusion/trainers/base_wan_trainer.py
```diff
@@ -178,9 +227,11 @@ def start_training(self):
     with nn_partitioning.axis_rules(self.config.logical_axis_rules):
       pipeline, opt_state, step = self.checkpointer.load_checkpoint()
     restore_args = {}
-    if opt_state and step:
-      restore_args = {"opt_state": opt_state, "step": step}
+    if opt_state is not None:
+      restore_args["opt_state"] = opt_state
       del opt_state
+    if step is not None:
+      restore_args["step"] = step
```
Perseus14, line 234: Could we handle weights-only checkpoints explicitly? This advances the training loop while Adam and its LR schedule restart at zero, making the actual and logged learning rates inconsistent. We should either reset the loop too or define how the schedule continues without optimizer state.

src/maxdiffusion/trainers/base_wan_trainer.py
```diff
@@ -261,21 +312,27 @@ def training_loop(self, pipeline, optimizer, learning_rate_scheduler, train_data
...
       state = jax.lax.with_sharding_constraint(state, state_spec)
       state_shardings = nnx.get_named_sharding(state, mesh)
-      if jax.process_index() == 0 and restore_args:
-        max_logging.log("--- Optimizer State Sharding Spec (opt_state) ---")
-        pretty_string = pprint.pformat(state_spec.opt_state, indent=4, width=60)
-        max_logging.log(pretty_string)
-        max_logging.log("------------------------------------------------")
+      if restore_args.get("step") is not None:
+        step = restore_args.get("step", 0)
+        state = state.replace(step=step)
```
Perseus14: Could we restore the saved `TrainState.step` and resume the loop from it? Checkpoint `100` is saved after `101` updates, but this restores step `100` while optimizer counters are at `101`. The loop also repeats iteration `100`. For legacy checkpoints, `checkpoint_index + 1` matches the existing save convention.

Perseus14 (commented):
Thanks for the fix @ninatu! Added a few comments, PTAL!

src/maxdiffusion/trainers/base_wan_trainer.py
ninatu: Agreed, weights-only now don't init the step (resets the loop to 0) and print a warning.

src/maxdiffusion/trainers/base_wan_trainer.py
ninatu: Good catch! fixed by resuming at N+1 for both `start_step` and `TrainState.step`.
The alternative would be returning `TrainState.step` from `self.checkpointer.load_checkpoint`
instead of the `step`, but that felt confusing given we pass `step` explicitly in
`checkpointer.save_checkpoint(step, ...)`, the two would mean different things. Let me know
if you disagree.

src/maxdiffusion/trainers/base_wan_trainer.py
```diff
@@ -256,26 +307,51 @@ def eval(self, mesh, eval_rng_key, step, p_eval_step, state, scheduler_state, wr
...
+          # and the loop starts at step 0, so it already reflects N+1 applied
+          # updates. Resume at N+1 so that iteration N is not re-run, and so
+          # that TrainState.step stays aligned with the optimizer's internal
+          # step count, which actually drives the LR schedule.
+          state = state.replace(step=step + 1)
+          start_step = step + 1
+          max_logging.log(f"Resuming training from step {start_step}")
+      else:
+        # Weights-only: no optimizer state means no schedule position.
+        # Resetting to 0 keeps the applied LR and the logged LR consistent.
```
Perseus14, line 346: Resetting the loop to zero works! One minor thing to note, as the existing checkpoint folder already contains specific steps checkpoint, so the new training run will skip saving the checkpoints at those steps. We might need to ensure that the new checkpoints are saved to a different folder.

Very minor edge case

ninatu: Good point. I updated the logic so that when resuming from weights-only, it automatically appends `_from_weights_<step>` to the run_name and re-initializes the checkpointer. This redirects all new checkpoints and TensorBoard logs into a fresh subfolder so they don't overlap.

And another point, I also had to include a fix for the flaky test_make_dreambooth_train_iterator test that  started to fail (I'm not sure why nobody got the same problem before). It was a race condition where parallel test workers were colliding over downloads to a shared, hardcoded /tmp/ directory. I updated the test to use an isolated tempfile.TemporaryDirectory() instead.

